PR #1, opened 2016-03-03
fix compilation errors because of updated dependencies

lrbalt:
because of updates of libc and bitflags, these two changes were needed to make the lib compile


diwic:
Thanks! Merged for now; but I added poll support to the libc crate a while ago, so maybe I should update the poll api to use that instead.

lrbalt:
always good to migrate code to lib supported by someone else :-)

PR #2, opened 2016-03-13
change test to use first pcm instead of hardcode hw:0

lrbalt:
My setup does not have hw:0


diwic:
Ok, but probably better to loop through all PCMs rather than just the first one.

lrbalt:
I see that you have already done this. Nice!

PR #4, opened 2016-03-17
implement basic iterator for simple mixer api

lrbalt:
@diwic could you review this first attempt? Am I handling unsafe calls correctly? 

Not meant for merging, just to get your review :-)

### DIFF
--- a/src/mixer.rs
+++ b/src/mixer.rs
@@ -263,13 +263,18 @@ impl<'a> Selem<'a> {
     ///
     /// # Example
     /// ```ignore
-    /// let db_value = selem.playback_volume_decibel(SelemChannelId::FrontLeft as i32).unwrap() as f32 / 100.0;
+    /// let db_value = selem.get_playback_volume_decibel(SelemChannelId::FrontLeft as i32).unwrap() as f32 / 100.0;
     /// ```
-    pub fn ask_playback_vol_decibel(&self, channel: i32) -> Result<i64> {
-        let mut decibel_value: i64 = 0;
+    pub fn get_playback_vol_decibel(&self, channel: i32) -> Result<i64> {
         self.get_playback_volume(channel)
-            .and_then(|volume| acheck!(snd_mixer_selem_ask_playback_vol_dB (self.1.handle, volume, &mut decibel_value)))
-            .and_then(|_| Ok(decibel_value))
+            .and_then(|volume| self.ask_playback_vol_decibel(volume))
+    }
+
+    /// Asks alsa to convert playback volume to decibels
+    pub fn ask_playback_vol_decibel(&self, volume: i64) -> Result<i64> {
+        let mut decibel_value: i64 = 0;
+        acheck!(snd_mixer_selem_ask_playback_vol_dB (self.1.handle, volume, &mut decibel_value))
+            .and_then(|_| Ok(decibel_value) )
     }
 
     pub fn get_capture_volume(&self, channel: i32) -> Result<i64> {
@@ -281,12 +286,17 @@ impl<'a> Selem<'a> {
     ///
     /// # Example
     /// ```ignore
-    /// let db_value = selem.capture_volume_decibel(SelemChannelId::FrontLeft as i32).unwrap() as f32 / 100.0;
+    /// let db_value = selem.get_capture_volume_decibel(SelemChannelId::FrontLeft as i32).unwrap() as f32 / 100.0;
     /// ```
-    pub fn ask_capture_vol_decibel(&self, channel: i32) -> Result<i64> {
-        let mut decibel_value: i64 = 0;
+    pub fn get_capture_vol_decibel(&self, channel: i32) -> Result<i64> {
         self.get_capture_volume(channel)
-            .and_then(|volume| acheck!(snd_mixer_selem_ask_capture_vol_dB (self.1.handle, volume, &mut decibel_value)))
+            .and_then(|volume| self.ask_capture_vol_decibel(volume))
+    }
+
+    /// Asks alsa to convert capture volume to decibels
+    pub fn ask_capture_vol_decibel(&self, volume: i64) -> Result<i64> {
+        let mut decibel_value: i64 = 0;
+        acheck!(snd_mixer_selem_ask_capture_vol_dB (self.1.handle, volume, &mut decibel_value))
             .and_then(|_| Ok(decibel_value))
     }
 
@@ -365,7 +375,7 @@ fn print_mixer_of_cards() {
                 for channel in 0..SelemChannelId::Last as i32 {
                     if selem.has_capture_channel(channel) { print!("{}:{}/{}dB ", channel,
                         match selem.get_capture_volume(channel) {Ok(v) => format!("{}",v).to_string(), Err(_) => "n/a".to_string()},
-                        match selem.ask_capture_vol_decibel(channel) {Ok(v) => format!("{}",v as f32 /100.0).to_string(), Err(_) => "n/a".to_string()}
+                        match selem.get_capture_vol_decibel(channel) {Ok(v) => format!("{}",v as f32 /100.0).to_string(), Err(_) => "n/a".to_string()}
                     );}
                 }
                 println!("");
@@ -387,7 +397,7 @@ fn print_mixer_of_cards() {
                         if selem.has_playback_channel(channel) { print!("{}:{}/{}dB ",
                             channel,
                             match selem.get_playback_volume(channel) {Ok(v) => format!("{}",v).to_string(), Err(_) => "n/a".to_string()},
-                            match selem.ask_playback_vol_decibel(channel) {Ok(v) => format!("{}",(v as f32) / 100.0).to_string(), Err(_) => "n/a".to_string()}
+                            match selem.get_playback_vol_decibel(channel) {Ok(v) => format!("{}",(v as f32) / 100.0).to_string(), Err(_) => "n/a".to_string()}
                         );}
                     }
                     println!("");
--- a/src/mixer.rs
+++ b/src/mixer.rs
@@ -1,3 +1,5 @@
+//! Mixer API - Simple Mixer API for mixer control
+//!
 use std::{ptr, mem};
 use std::ffi::CString;
 
@@ -6,7 +8,20 @@ use super::card;
 use super::error::*;
 
 const SELEM_ID_SIZE: usize = 64;
+const SND_MIXER_SCHN_LAST: i32 = 31;
 
+/// Iterator over available mixers of a card. This Iter wraps
+/// [snd_mixer_elem_t](http://www.alsa-project.org/alsa-doc/alsa-lib/group___mixer.html) and
+/// uses [snd_mixer_t](http://www.alsa-project.org/alsa-doc/alsa-lib/group___mixer.html)
+///
+/// # Example
+/// ```
+/// use alsa::card;
+/// let card = card::Card::new(0);  // use soundcard 0
+/// for mixer in Iter::new(card).unwrap().map(|m| m.unwrap()) {
+///   println!("Card #{}: {} ({})", card.get_index(), card.get_name().unwrap(), card.get_longname().unwrap());
+/// }
+/// ```
 #[derive(Debug)]
 pub struct Iter {
     handle: *mut alsa::snd_mixer_t,
@@ -14,6 +29,7 @@ pub struct Iter {
 }
 
 impl Iter {
+    /// Creates a new iterator for a specific card using the cards using hw name, i.e. `hw:0`
     pub fn new(c: card::Card) -> Result<Iter> {
         let card = &CString::new(format!("hw:{}", c.get_index())).unwrap();
         let mut mixer_handle = ptr::null_mut();
@@ -30,6 +46,7 @@ impl Iter {
     }
 }
 
+/// Closes mixer and frees used alsa resources
 impl Drop for Iter {
     fn drop(&mut self) {
         unsafe { alsa::snd_mixer_close(self.handle) };
@@ -51,27 +68,31 @@ impl Iterator for Iter {
         if elem.is_null() {
             None
         } else {
+            // remember last used handle to use it for snd_mixer_elem_next at next call of next()
             self.previous = Some(elem.handle);
             Some(Ok(elem))
         }
     }
 }
 
+// use fixed size structure for selem_id
 pub struct SelemId([u8; SELEM_ID_SIZE]);
 
+/// Creates a new `selem_id` using `SelemId` of hardcoded size. This size is checked with `snd_mixer_selem_id_sizeof`
 pub fn selem_id_new() -> Result<SelemId> {
     assert!(unsafe { alsa::snd_mixer_selem_id_sizeof() } as usize <= SELEM_ID_SIZE);
     Ok(SelemId(unsafe { mem::zeroed() }))
 }
 
+/// Convert SelemId into ``*mut *mut snd_mixer_selem_id_t` that the alsa call needs. See [snd_mixer_selem_id_t](http://www.alsa-project.org/alsa-doc/alsa-lib/group___simple_mixer.html)
 #[inline]
 pub fn selem_id_ptr(a: &SelemId) -> *mut alsa::snd_mixer_selem_id_t {
     a.0.as_ptr() as *const _ as *mut alsa::snd_mixer_selem_id_t
 }
 
+/// Mixer wraps [snd_mixer_selem_id_t](http://www.alsa-project.org/alsa-doc/alsa-lib/group___simple_mixer.html)
 pub struct Mixer {
     handle: *mut alsa::snd_mixer_elem_t,
-    // selem_id: *mut alsa::snd_mixer_selem_id_t
     selem_id: SelemId
 }
 
@@ -79,6 +100,7 @@ impl Mixer {
     pub fn new(mixer_handle: *mut alsa::snd_mixer_elem_t) -> Mixer {
         let sid = selem_id_new().unwrap();
 
+        // check for null pointer
         if mixer_handle != 0 as *mut alsa::snd_mixer_elem_t {
             unsafe { alsa::snd_mixer_selem_get_id(mixer_handle, selem_id_ptr(&sid)) };
         }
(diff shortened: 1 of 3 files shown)


diwic:
Thanks for working on it! My high-level objection is that it's not following the ALSA names clearly and closely enough. It should be obvious what alsa-lib call they wrap. So we need to start with

    // in mixer.rs
    pub struct Mixer(*mut snd_mixer_t);
    pub struct Elem<'a>(*mut snd_mixer_elem_t, &'a Mixer);
    pub struct Iter<'a>(*mut snd_mixer_elem_t, &'a Mixer); // For iterating over elements in a mixer

Not sure if it makes sense to wrap a mixerchannel, isn't it easier just to supply the parameter at every call like alsa-lib does? And a mixer channel is probably not an i32, but an `pub enum SelemChannelId`.

You asked about the unsafe stuff, I don't see any big mistakes there.

diwic:
Ok, since you needed some more hand-holding w r t the design, I looked a little bit more and came up with this stub:

    pub struct Mixer(*mut snd_mixer_t);
    
    #[derive(Copy, Clone)]
    pub struct Elem<'a>(*mut snd_mixer_elem_t, &'a Mixer);
    
    #[derive(Copy, Clone)]
    pub struct Selem<'a>(Elem<'a>);
    
    #[derive(Copy, Clone)]
    pub struct Iter<'a>(*mut snd_mixer_elem_t, &'a Mixer);
    
    impl<'a> Iterator for Iter<'a> {
        type Item = Elem<'a>;
    }

    #[derive(Copy, Clone)]
    pub struct SelemId([u8; SELEM_ID_SIZE]);
    
    #[derive(Copy, Clone)]
    pub enum SelemChannelId { 
        FrontLeft = alsa::SND_MIXER_SCHN_FRONT_LEFT as isize,
        /* ... */
    };

For most use cases Elem and Selem are one and the same, but sometimes there are Elems which are not Selems. So I think we should have both to be future-proof, and I could add support for other mixer classes at some later point in time.

The thing to be careful about here is the relationship between snd_mixer_t and the snd_mixer_elem_t it owns internally - that's why our Elem needs to hold a reference to Mixer.

lrbalt:
Thanks for the feedback. I'll take another shot at this.

lrbalt:
Ok, I've used the design you provided. I have kept naming of methods closer to alsa api and suppressed my urge to use more OO-principles :-)

I placed all selem methods in `Selem`, so `Elem` is just holding the `snd_mixer_elem_t`.

src/mixer.rs
```diff
@@ -0,0 +1,428 @@
...
+}
+
+/// Closes mixer and frees used resources
+impl Drop for Mixer {
+    fn drop(&mut self) {
+        unsafe { alsa::snd_mixer_close(self.0) };
+    }
+}
+
+/// Wraps [snd_mixer_selem_id_t](http://www.alsa-project.org/alsa-doc/alsa-lib/group___simple_mixer.html)
```
diwic: Typo

src/mixer.rs
```diff
@@ -0,0 +1,428 @@
...
+pub struct Mixer(*mut alsa::snd_mixer_t);
+
+impl Mixer {
+    /// Creates a new iterator for a specific card using the cards index and
+    /// using hw name, i.e. `hw:0`
+    pub fn new(c: &card::Card) -> Result<Mixer> {
+        let card = &CString::new(format!("hw:{}", c.get_index())).unwrap();
+        let mut mixer_handle = ptr::null_mut();
+        try!(acheck!(snd_mixer_open(&mut mixer_handle, 0)));
+        try!(acheck!(snd_mixer_attach(mixer_handle, card.as_ptr())));
```
diwic: Need to put handle into Mixer already here - as long as you've done snd_mixer_open, you must do snd_mixer_close or you'll leak the handle

src/mixer.rs
```diff
@@ -0,0 +1,428 @@
...
+
+const SELEM_ID_SIZE: usize = 64;
+
+/// wraps [snd_mixer_t](http://www.alsa-project.org/alsa-doc/alsa-lib/group___mixer.html)
+pub struct Mixer(*mut alsa::snd_mixer_t);
+
+impl Mixer {
+    /// Creates a new iterator for a specific card using the cards index and
+    /// using hw name, i.e. `hw:0`
+    pub fn new(c: &card::Card) -> Result<Mixer> {
```
diwic: Better to provide string as input; ioplugs can have mixers and they do not have the syntax hw:<number>

src/mixer.rs
```diff
@@ -0,0 +1,428 @@
...
+/// Wraps [snd_mixer_selem_id_t](http://www.alsa-project.org/alsa-doc/alsa-lib/group___simple_mixer.html)
+#[derive(Copy, Clone)]
+pub struct Elem<'a>{
+    handle: *mut alsa::snd_mixer_elem_t,
+    mixer: &'a Mixer
+}
+
+impl<'a> Elem<'a> {
+    /// Checks if this element is null by checking if the handle is null
+    pub fn is_null(&self) -> bool {
```
diwic: Remove this function; Elem should always contain a non-null pointer

src/mixer.rs
```diff
@@ -0,0 +1,428 @@
...
+    pub fn new(elem: Elem) -> SelemId {
+        // Create empty selem_id and fill from mixer
+        let sid = SelemId::empty();
+        unsafe { alsa::snd_mixer_selem_get_id(elem.handle, sid.as_ptr()) };
+        sid
+    }
+
+    /// Returns an empty (zeroed) SelemId. This id is not a useable id and need to be initialized
+    /// like `SelemId::new()` does
+    pub fn empty() -> SelemId {
```
diwic: We should not have unusable id's, instead have a function that takes name and index and creates an SelemId from that

src/mixer.rs
```diff
@@ -0,0 +1,428 @@
...
+    /// See [snd_mixer_selem_id_t](http://www.alsa-project.org/alsa-doc/alsa-lib/group___simple_mixer.html)
+    #[inline]
+    pub fn as_ptr(&self) -> *mut alsa::snd_mixer_selem_id_t {
+        self.0.as_ptr() as *const _ as *mut alsa::snd_mixer_selem_id_t
+    }
+}
+
+/// Wraps an Elem as a Selem
+// #[derive(Copy, Clone)]
+pub struct Selem<'a>(SelemId, Elem<'a>);
```
diwic: This is incorrect, the SelemId is stored internally inside the Elem, we shouldn't store a duplicate

src/mixer.rs
```diff
@@ -0,0 +1,428 @@
...
+
+    /// Creates a Selem by lookung for a specific selem by name given a mixer (of a card)
+    pub fn find_by_name(mixer: &'a Mixer, name: &str) -> Result<Selem<'a>> {
+        let sid = SelemId::empty();
+        unsafe { alsa::snd_mixer_selem_id_set_index(sid.as_ptr(), 0) };
+        unsafe { alsa::snd_mixer_selem_id_set_name(sid.as_ptr(), CString::new(name).unwrap().as_ptr()) };
+        let elem = Elem { handle: unsafe { alsa::snd_mixer_find_selem(mixer.0, sid.as_ptr()) }, mixer: mixer};
+
+        if elem.is_null() {
+            Err(Error::new(Some(stringify!("snd_mixer_find_selem").into()), -1 as ::libc::c_int))
```
diwic: stringify?

src/mixer.rs
```diff
@@ -0,0 +1,428 @@
...
+pub struct Selem<'a>(SelemId, Elem<'a>);
+
+impl<'a> Selem<'a> {
+    /// Creates a Selem by wrapping `elem`.
+    pub fn new(elem: Elem<'a>) -> Selem<'a> {
+        Selem(SelemId::new(elem), elem)
+    }
+
+    /// Creates a Selem by lookung for a specific selem by name given a mixer (of a card)
+    pub fn find_by_name(mixer: &'a Mixer, name: &str) -> Result<Selem<'a>> {
```
diwic: What's wrong with having a find_selem function on Mixer instead? Since this is basically a wrapper around snd_mixer_find_selem

src/mixer.rs
```diff
@@ -0,0 +1,428 @@
...
+
+        if elem.is_null() {
+            Err(Error::new(Some(stringify!("snd_mixer_find_selem").into()), -1 as ::libc::c_int))
+        } else {
+            Ok(Selem::new(elem))
+        }
+    }
+
+    /// returns the elem of this selem
+    pub fn get_elem(&self) -> Elem<'a> {
```
diwic: For some elegance, we could implement Deref instead

src/mixer.rs
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    pub fn has_playback_channel(&self, channel: i32) -> bool {
+        unsafe { alsa::snd_mixer_selem_has_playback_channel(self.1.handle, channel) > 0 }
+    }
+
+    /// Gets name from snd_mixer_selem_channel_name
+    pub fn channel_name(&self, channel: i32) -> Result<String> {
+        let c = unsafe { alsa::snd_mixer_selem_channel_name(channel) };
+        from_const("snd_mixer_selem_channel_name", c).and_then(|s| Ok(s.to_string()))
```
diwic: This is a weird one, it does not need to take self. I think the alsa-lib function is misnamed (should belong to channel_id instead)?

src/mixer.rs
```diff
@@ -0,0 +1,428 @@
...
+    RearLeft    = alsa::SND_MIXER_SCHN_REAR_LEFT as isize,
+    RearRight   = alsa::SND_MIXER_SCHN_REAR_RIGHT as isize,
+    FrontCenter = alsa::SND_MIXER_SCHN_FRONT_CENTER as isize,
+    Woofer      = alsa::SND_MIXER_SCHN_WOOFER as isize,
+    SideLeft    = alsa::SND_MIXER_SCHN_SIDE_LEFT as isize,
+    SideRight   = alsa::SND_MIXER_SCHN_SIDE_RIGHT as isize,
+    RearCenter  = alsa::SND_MIXER_SCHN_REAR_CENTER as isize,
+    Last        = alsa::SND_MIXER_SCHN_LAST as isize,
+    // Mono        = alsa::SND_MIXER_SCHN_MONO as isize,
+}
```
diwic: When there are duplicate constants (such as Front_left and mono), we can provide a mono() function on SelemChannelId instead

src/mixer.rs
```diff
@@ -0,0 +1,428 @@
...
+
+impl SelemId {
+    /// Creates a new SelemId` of hardcoded size SELEM_ID_SIZE.
+    /// This size is checked against `snd_mixer_selem_id_sizeof`
+    pub fn new(elem: Elem) -> SelemId {
+        // Create empty selem_id and fill from mixer
+        let sid = SelemId::empty();
+        unsafe { alsa::snd_mixer_selem_get_id(elem.handle, sid.as_ptr()) };
+        sid
+    }
```
diwic: Where is set and get for name and index?

src/mixer.rs
```diff
@@ -0,0 +1,428 @@
...
+    pub fn empty() -> SelemId {
+        assert!(unsafe { alsa::snd_mixer_selem_id_sizeof() } as usize <= SELEM_ID_SIZE);
+        // Create empty selem_id and fill from mixer
+        SelemId(unsafe { mem::zeroed() })
+    }
+
+    /// Convert SelemId into ``*mut *mut snd_mixer_selem_id_t` that the alsa call needs.
+    /// See [snd_mixer_selem_id_t](http://www.alsa-project.org/alsa-doc/alsa-lib/group___simple_mixer.html)
+    #[inline]
+    pub fn as_ptr(&self) -> *mut alsa::snd_mixer_selem_id_t {
```
diwic: I don't want to expose ffi pointers directly, so this function should not be public

diwic:
Ok, I have commented on the PR now. It looks significantly better, but you still have a tendency to wrap more than a single alsa-lib call in your functions. Try to avoid that - you might make your use case slightly more simple, but you break many other use cases. E g, your `ask_playback_vol_decibel` wraps `snd_mixer_selem_ask_playback_vol_dB` in a way that you can now only use it to find the dB value of the *current* volume, therefore you no longer support the use case of wanting to find the dB value of some other volume.

src/mixer.rs
```diff
@@ -0,0 +1,428 @@
...
+pub struct SelemId([u8; SELEM_ID_SIZE]);
+
+impl SelemId {
+    /// Creates a new SelemId` of hardcoded size SELEM_ID_SIZE.
+    /// This size is checked against `snd_mixer_selem_id_sizeof`
+    pub fn new(elem: Elem) -> SelemId {
+        // Create empty selem_id and fill from mixer
+        let sid = SelemId::empty();
+        unsafe { alsa::snd_mixer_selem_get_id(elem.handle, sid.as_ptr()) };
+        sid
```
diwic: This function should not be SelemId::new, it should be Selem::get_id()

src/mixer.rs
```diff
@@ -0,0 +1,428 @@
...
+
+    /// returns the elem of this selem
+    pub fn get_elem(&self) -> Elem<'a> {
+        self.1
+    }
+
+    pub fn get_name(&self) -> Result<String> {
+        let c = unsafe { alsa::snd_mixer_selem_id_get_name(self.0.as_ptr()) };
+        from_const("snd_mixer_selem_id_get_name", c).and_then(|s| Ok(s.to_string()))
+    }
```
diwic: Should move to SelemId. No need to do the to_string part, we can return a pointer to the SelemId and skip an allocation

src/mixer.rs
lrbalt: Perhaps I don't understand the difference or similarity between 'Selem' and 'Elem', but 'Elem' wraps 'snd_mixer_elem_t' and 'Selem' wraps 'SelemId' which is of type 'snd_mixer_selem_id_t'. 

Are these the same?

src/mixer.rs
lrbalt: What do you propose. Shall we follow alsa naming (although perhaps misnamed) or move this to SelemChannelId?

src/mixer.rs
lrbalt: corrected

src/mixer.rs
lrbalt: I create a Mixer on line 20 now

src/mixer.rs
lrbalt: done

src/mixer.rs
lrbalt: removed. Was left over from earlier design

src/mixer.rs
lrbalt: copy-and-paste error :-) Removed

src/mixer.rs
lrbalt: moved to Mixer

src/mixer.rs
lrbalt: Added Deref trait implementation

src/mixer.rs
lrbalt: Done

src/mixer.rs
lrbalt: I had both in `Selem`. Now moved from `Selem` to `SelemId`

src/mixer.rs
lrbalt: removed `pub`

src/mixer.rs
lrbalt: renamed

src/mixer.rs
lrbalt: moved both `get_id` and `get_name` to `SelemId`

lrbalt:
"you still have a tendency to wrap more than a single alsa-lib call in your functions." 

Are you referring to `Mixer::new` for example? I think it is now convenience to get a working mixer from `Mixer::new`. But I can make four different functions for `open`, `attach`, `selem_register` and `load`. Perhaps leave `new` as a convenience function?

Could do the same for `Mixer::find_selem` and `ask_playback_vol_decibel`. It will make it more difficult to get you going with alsa though. Or do you like to get a convenience layer on top of alsa-rs someday?

diwic:
> But I can make four different functions for open, attach, selem_register and load. Perhaps leave new as a convenience function?

Yes.

> Could do the same for Mixer::find_selem and ask_playback_vol_decibel. It will make it more difficult to get you going with alsa though. 

It depends on where you come from. If you're used to using ALSA but new to Rust, then the 1-1 mappers would be the easiest to get started with. It's also a matter of consistency, that the Mixer module has the same design philosophy (see Readme.md) as the rest of the crate. And third, if you read alsa-lib documentation that mentions C functions, it should be easy to find the corresponding Rust functions.

I'm okay with leaving some convenience functions in, such as Mixer::new. But the most important stuff is to provide Rust's safety, and as thin wrappers on top of that as possible.

Btw, is there a reason why you have an Iter::new function? Having an Mixer::iter function seems more rustic to me.

lrbalt:
I hope we are close to merging now. I think I addressed all your comments

diwic:
> I hope we are close to merging now. I think I addressed all your comments

If that's the way you feel about it, how about I take your code as it is and continue working from there?

I briefly looked at the test code too, now - I was wondering what to do with the get_and_set tests, since most people won't have a hw:2 card with Speaker and Mic elements in them? Maybe I should just comment them out?

src/mixer.rs
diwic: The alsa API is a bit over-engineered here. If you like thinking OO, then think of `Selem` as a subclass of `Elem`. `SelemId` is just a name and an index (which uniquely maps to a `Selem`), and `Selem` internally contains both a `SelemId` and per-channel information about current volume, min/max values, etc.

lrbalt:
I think it is a good idea if you take it from here. At this point I'm asking you what to do, so it is probably faster / efficient if you do the changes right away.

Wrt the test code, I was working on finding the first card with a playback mixer selem, but could not get the code working, so I reverted to a poor-mans solution by hard coding a playback selem of my soundcard 2. You could comment it out or create a small finder for a playback and capture selem.

diwic:
Okay, I've now squashed all your commits into one, and provided another commit on top which fixes up a lot of remaining things.

What's not supported yet is enums (snd_mixer_selem_get_enum_*). I'm a bit unsure about what to do with your can_playback function, should that return true or false for a playback enum?

lrbalt:
Thanks for merging! I'd suggest waiting with the enum stuff and add it on demand :-)

I got the can_playback function from the amixer source. It is a convenience function, so you could remove it or move it somewhere else.

PR #8, opened 2016-05-20
add capture and playback switches

lrbalt:
I needed to be able to unmute line-in. It was missing from selem struct.


diwic:
Looks good, thanks! Although get_*_switch should probably be added too, for completeness.

PR #9, opened 2016-05-20
add get_{playback,capture}_switch too

lrbalt:


lrbalt:
as requested :-)

diwic:
:+1:

PR #10, opened 2016-08-07
add setting volume in decibel

lrbalt:
I've added [snd_mixer_selem_set_playback_dB](http://www.alsa-project.org/alsa-doc/alsa-lib/group___simple_mixer.html#ga8bfb52575e4bb06deb30799bc39c0768)

This function uses `dir` to `select direction (-1 = accurate or first bellow, 0 = accurate, 1 = accurate or first above)`. I've added an enum for this called `AccuracyDirection` but there is also a `ValueOr` enum in `lib.rs`. Should I use ValueOr for this function too?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -91,7 +91,6 @@ pub use poll::PollDescriptors as PollDescriptors;
 
 pub mod mixer;
 pub use mixer::Mixer as Mixer;
-pub use mixer::AccuracyDirection as AccuracyDirection;
 
 mod io;
 pub use io::Output;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -91,6 +91,7 @@ pub use poll::PollDescriptors as PollDescriptors;
 
 pub mod mixer;
 pub use mixer::Mixer as Mixer;
+pub use mixer::AccuracyDirection as AccuracyDirection;
 
 mod io;
 pub use io::Output;
--- a/src/mixer.rs
+++ b/src/mixer.rs
@@ -3,9 +3,10 @@
 use std::ffi::{CStr, CString};
 use std::{ptr, mem, fmt};
 use std::ops::Deref;
-use libc::{c_long};
+use libc::{c_long, c_int};
 
 use alsa;
+use super::ValueOr;
 use super::error::*;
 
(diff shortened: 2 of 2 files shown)


diwic:
Please use `ValueOr`. And the function name should match alsa-libs (i e, should be "set_playback_db", not "set_playback_volume_decibel". Looks good otherwise!

diwic:
Hi, I just added a millibel wrapper to clarify that ALSA returns and sets values in millibels instead of decibels (I believe this is true even when not explicitly mentioned in the docs, I checked in the source code). Could you adjust your new function to use that as well? Thanks.

lrbalt:
Other functions use `decibel` too, because `dB` cannot be used (camelcase) without warning. 

Shall I rename all uses of `decibel` to `db` (lowercase) where ALSA uses `dB`?

I will migrate to the millibel wrapper too.

diwic:
> Shall I rename all uses of decibel to db (lowercase) where ALSA uses dB?

I already did so in my commit, unless I missed something?

 > I will migrate to the millibel wrapper too.

Thanks. Btw, do you prefer `MilliBel` over the current non-camelcase, but shorter, `mb`?  I couldn't really make up my mind of which one was better. (`MB` and `Mb` could both mean "Mega" instead of "Milli" so I didn't want to use them.)

diwic:
http://mailman.alsa-project.org/pipermail/alsa-devel/2016-August/111743.html

lrbalt:
The PR is changed to use `mb` and `ValueOr`.

I did not change ValueOr to conform to the patch you posted on ALSA. Should we wait until it is accepted?

lrbalt:
I'm not fond of mb. It is not as intuitive as db is for decibel. I prefer decibel of db and millibel over mb for readability. But I understand that you want to keep as close to ALSA API as possible.

Since mb is not in ALSA API, perhaps you can use `MilliBel` for the wrapper and leave `db` for `decibel`?

diwic:
Renamed mb to MilliBel now. Let's wait for upstream; if accepted then maybe we should not use ValueOr but a simple bool instead for this particular function.

PR #23, opened 2017-04-01
Add ability for Event to reference external buffer

Boddlnagg:
This is the first version of a PR that allows `seq::Event` to carry a reference to an external buffer for variable length data (esp. SysEx events). Without this, there is no way to use `encode` for SysEx messages and then send the message without creating an additional copy of the buffer. I removed the `encode_cb` variant, because it makes no sense: For the encoded event to be sent, it *must* contain a reference to the actual buffer. Actually, having a SysEx message that references an empty buffer (which was previously possible by returning `None` from the callback) is probably an error. ALSA forwards such an event unchecked, but the receiving application might not be able to deal with it.

I don't consider th e added lifetime for `Event<'a>` a big usability issue. IMHO it actually makes the API nicer compared to the callback variant. `encode` now takes the encoder (`MidiEvent`) by unique reference, so that the encoder can't be messed with while the event still references its buffer. This statically enforces that no method listed in the [documentation for `encode`](http://www.alsa-project.org/alsa-doc/alsa-lib/group___m_i_d_i___event.html#gae44531e8091bdd6e97a5b70e7b81ea10) is called and the underyling buffer within the encoder remains unchanged while the event is alive.

The following changes could be added on top, but I didn't do it yet because I wanted to get feedbackfirst:
- `event_input` could be changed in a similar way to `encode`, coupling the returned event's lifetime to the `Seq` instance by taking `&mut Seq`. In this case, the input buffer of the sequencer instance is directly referenced, and it's not very well documented under what circumstances that buffer might change. Not allowing any method call on `Seq` while the event is alive should be enough, however, but might be too much, because e.g. you wouldn't be able to call `event_output` directly after `event_input` to forward the event to another port.

- It is desirable to be able to create an `Event<'a>` directly from a `&'a [u8]`, i.e. a SysEx message that references a buffer managed by the user, without creating a copy. (Update: I have added another commit (the second one) to implement this. Feel free to reject that commit.)

Update: I have also added a third commit that fixes a preexisting bug that is not directly related to my changes.

### DIFF
--- a/src/seq.rs
+++ b/src/seq.rs
@@ -4,7 +4,7 @@ use libc::{c_uint, c_int, c_short, c_uchar, c_void, c_long, size_t, pollfd};
 use super::error::*;
 use alsa;
 use super::{Direction, poll};
-use std::{ptr, fmt, mem, slice, time};
+use std::{ptr, fmt, mem, slice, time, cell};
 use std::ffi::CStr;
 use std::borrow::Cow;
 
@@ -40,7 +40,11 @@ struct EvExtPacked {
 }
 
 /// [snd_seq_t](http://www.alsa-project.org/alsa-doc/alsa-lib/group___sequencer.html) wrapper
-pub struct Seq(*mut alsa::snd_seq_t);
+///
+/// To access the functions ``event_input`, `event_input_pending` and `set_input_buffer_size`,
+/// you first have to obtain an instance of `Input` by calling `input()`. Only one instance of
+/// `Input` may exist at any time for a given `Seq`.
+pub struct Seq(*mut alsa::snd_seq_t, cell::Cell<bool>);
 
 unsafe impl Send for Seq {}
 
@@ -49,6 +53,10 @@ impl Drop for Seq {
 }
 
 impl Seq {
+    fn check_has_input(&self) {
+        if self.1.get() { panic!("No additional Input object allowed")}
+    }
+
     /// If name is None, "default" will be used.
     pub fn open(name: Option<&CStr>, dir: Option<Direction>, nonblock: bool) -> Result<Seq> {
         let n2 = name.unwrap_or(unsafe { CStr::from_bytes_with_nul_unchecked(b"default\0") });
@@ -60,7 +68,7 @@ impl Seq {
             Some(Direction::Capture) => SND_SEQ_OPEN_INPUT,
         };
         acheck!(snd_seq_open(&mut h, n2.as_ptr(), streams, mode))
-            .map(|_| Seq(h))
+            .map(|_| Seq(h, cell::Cell::new(false)))
     }
 
     pub fn set_client_name(&self, name: &CStr) -> Result<()> {
@@ -143,27 +151,6 @@ impl Seq {
         acheck!(snd_seq_event_output_direct(self.0, &mut e.0)).map(|q| q as u32)
     }
 
-    /// Note: this function is a non-allocating version of event_input.
-    ///
-    /// In case of sysex or other custom data length message, the supplied callback will be called with the buffer.
-    /// The return value of the callback is only relevant if you intend to call get_data::<Vec<u8>>() on the returned Event
-    /// or the event is passed back unchanged to ALSA for output.
-    pub fn event_input_cb<F: FnOnce(EventType, &[u8]) -> Option<Vec<u8>>>(&self, f: F) -> Result<Event<'static>> {
-        let mut z = ptr::null_mut();
-        try!(acheck!(snd_seq_event_input(self.0, &mut z)));
-        unsafe { Event::extract(&mut *z, "snd_seq_event_input", f) }
-    }
-
-    /// Note: this function will allocate in case of sysex or other custom data length messages.
-    /// If this is undesired, you can use `event_input_cb` instead.
-    pub fn event_input(&self) -> Result<Event<'static>> {
-        self.event_input_cb(|_,buf| Some(buf.to_vec()))
-    }
-
-    pub fn event_input_pending(&self, fetch_sequencer: bool) -> Result<u32> {
-        acheck!(snd_seq_event_input_pending(self.0, if fetch_sequencer {1} else {0})).map(|q| q as u32)
-    }
-
     pub fn get_queue_tempo(&self, q: i32) -> Result<QueueTempo> {
         let value = try!(QueueTempo::new());
         acheck!(snd_seq_get_queue_tempo(self.0, q as c_int, value.0)).map(|_| value)
@@ -177,6 +164,42 @@ impl Seq {
     pub fn alloc_queue(&self) -> Result<i32> { acheck!(snd_seq_alloc_queue(self.0)).map(|q| q as i32) }
     pub fn alloc_named_queue(&self, n: &CStr) -> Result<i32> {
         acheck!(snd_seq_alloc_named_queue(self.0, n.as_ptr())).map(|q| q as i32)
+    }
+
+    pub fn input<'a>(&'a self) -> Input<'a> {
+        Input::new(self)
+    }
+}
+
+pub struct Input<'a>(&'a Seq);
+
+impl<'a> Drop for Input<'a> {
+    fn drop(&mut self) { (self.0).1.set(false) }
+}
+
+impl<'a> Input<'a> {
+    fn new(s: &'a Seq) -> Input<'a> {
+        s.check_has_input();
+        s.1.set(true);
+        Input(s)
+    }
+
+    pub fn event_input<'b>(&'b mut self) -> Result<Event<'b>> {
+        // The returned event might reference the input buffer of the `Seq`.
+        // Therefore we mutably borrow the `Input` structure, preventing any
+        // other function call that might change the input buffer while the
+        // event is alive.
+        let mut z = ptr::null_mut();
+        try!(acheck!(snd_seq_event_input((self.0).0, &mut z)));
+        unsafe { Event::extract (&mut *z, "snd_seq_event_input") }
+    }
+
+    pub fn event_input_pending(&self, fetch_sequencer: bool) -> Result<u32> {
+        acheck!(snd_seq_event_input_pending((self.0).0, if fetch_sequencer {1} else {0})).map(|q| q as u32)
+    }
+
+    pub fn set_input_buffer_size(&self, size: u32)  -> Result<()> {
+        acheck!(snd_seq_set_input_buffer_size((self.0).0, size as size_t)).map(|_| ())
     }
 }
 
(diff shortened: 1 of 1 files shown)


diwic:
Before discussing the implementation details, could you explain why avoiding memory copies is so important to you?

Boddlnagg:
@diwic I think Rust is designed to allow zero-cost abstractions, and unnecessary copying should be avoided if there is a more efficient but still safe way. Not all copying is unnecessary, and actually a lot of what ALSA does internally is copying memory from user space to kernel space and back, but I think that a wrapper library should add more copying. If this library is meant to provide a thin wrapper for ALSA, then it seems counterproductive to me when the wrapper sometimes does an equal or greater amount of work than the underlying implementation. Admittedly, SysEx messages are probably quite a corner case, so you might argue that it's not worth making this one case more efficient by complicating the API in general, but as I said above, the `Event<'a>` version feels more like idiomatic Rust to me than the callback version.

diwic:
I had a look a snd_seq_event_input. Possible functions that can destroy ext data is: 
 * snd_seq_event_input
 * snd_seq_event_input_pending
 * snd_seq_event_input_set_buffer_size
 
If we want to tie the lifetime of `Event` then that should be done to these three functions.

Boddlnagg:
That's not so easy ... it would require splitting the `Seq` object in two parts where one would be responsible for these three functions and the other part would be responsible for the remaining functions. The problem is that those two parts would also need to be coupled to each other in some way ... I need to think about that a bit more, but it can be done in a separate PR.

diwic:
I do that for [PCM](https://github.com/diwic/alsa-rs/blob/master/src/pcm.rs#L96) today already, let's use the same tactic for seq then. 

I think you have a point about the zero-copy stuff, and after all, you're the only API user and if you like `Event<'a>` then fine :-)

.gitignore
```diff
@@ -0,0 +1,3 @@
+target/
```
diwic: These are unrelated changes.

src/seq.rs
```diff
@@ -141,16 +145,17 @@ impl Seq {
...
         let mut z = ptr::null_mut();
         try!(acheck!(snd_seq_event_input(self.0, &mut z)));
         unsafe { Event::extract(&mut *z, "snd_seq_event_input", f) }
     }
 
     /// Note: this function will allocate in case of sysex or other custom data length messages.
-    /// If this is undesired, you can use event_input_cb instead.
-    pub fn event_input(&self) -> Result<Event> {
+    /// If this is undesired, you can use `event_input_cb` instead.
+    pub fn event_input<'a, 'b>(&'a self) -> Result<Event<'b>> {
```
diwic: You're creating a lifetime out of thin air here. You really mean `Event<'static>`, right?

src/seq.rs
```diff
@@ -475,49 +480,97 @@ impl PortSubscribe {
...
+
 /// [snd_seq_event_t](http://www.alsa-project.org/alsa-doc/alsa-lib/structsnd__seq__event__t.html) wrapper
 ///
 /// Fields of the event is not directly exposed. Instead call `Event::new` to set data (which can be, e g, an EvNote).
 /// Use `get_type` and `get_data` to retreive data.
-pub struct Event(alsa::snd_seq_event_t, EventType, Option<Vec<u8>>);
+///
+/// The lifetime parameter refers to the lifetime of an associated external buffer that might be used for
+/// variable-length messages (e.g. SysEx).
+pub struct Event<'a>(alsa::snd_seq_event_t, EventType, EventExtraData<'a>);
```
diwic: How about `Option<Cow<'a, [u8]>>` instead of making a new `EventExtraData` struct?

src/seq.rs
```diff
@@ -475,49 +480,97 @@ impl PortSubscribe {
...
 
-impl Event {
-    pub fn new<D: EventData>(t: EventType, data: &D) -> Self {
-        let mut z = Event(unsafe { mem::zeroed() }, EventType::None, None);
+impl<'a> Event<'a> {
+    pub fn new<D: SetEventData>(t: EventType, data: D) -> D::Output {
+        let mut z = Event(unsafe { mem::zeroed() }, EventType::None, EventExtraData::None);
         z.1 = t;
         (z.0)._type = t as c_uchar;
+        let length_flags = match t {
```
diwic: This part should be refactored into its own function, and should probably be moved to ensure_buf?

src/seq.rs
```diff
@@ -475,49 +480,97 @@ impl PortSubscribe {
...
         debug_assert!(D::has_data(t));
-        data.set_data(&mut z);
-        z
+        data.set_data(z)
+    }
+
+    /// Consumes this event and returns an (otherwise unchanged) event where the externally referenced
+    /// buffer for variable length messages (e.g. SysEx) has been copied into the event.
+    /// The returned event has an arbitrary lifetime that is decoupled from the original buffer.
+    pub fn into_owned<'b>(self) -> Event<'b> {
```
diwic: Same here, this should be `into_static(self) -> Event<'static>`

Boddlnagg:
I haven't actually looked at the non-MIDI stuff so far ... going to do that, thanks for the hint! As far as `Event<'a>` is concerned, you're probably going to deal with `Event<'static>` most of the time (actually now that I think about it, [this](https://github.com/diwic/alsa-rs/pull/23/commits/fca6d8d43e48e76061f0454426db5abdc8d83b93#diff-f0a22d0ca5f8f67f38198c46ef26ad68R491) should probably return `Event<'static>` too, instead of a generic unrestrained `'b` lifetime.

src/seq.rs
```diff
@@ -601,23 +654,41 @@ impl fmt::Debug for Event {
     }
 }
 
-/// Low level methods to set/get data on an Event. Don't use these directly, use generic methods on Event instead.
-pub trait EventData {
+/// Low level method to check if event type supports a given data type. Don't use directly, use generic methods on Event instead.
+pub trait HasEventData {
```
diwic: I'm not really happy about the split into three traits and the associated type on SetEventData. If we instead had an extra `Event::new_borrowed` function for the particular case of borrowing `&[u8]`, would that help?

Boddlnagg: Yes, I think that's a reasonable alternative!

src/seq.rs
Boddlnagg: That would be equivalent, I think, but in my opinion the extra struct makes the intent clearer to the reader of the code. I'm fine with either way, since this is an implementation detail (the `EventExtraData` struct should never be exposed as part of the public API).

.gitignore
Boddlnagg: Yes, I'll create a separate PR for this. Or is there a reason that you did not have a `.gitignore`?

diwic: Just that the crate was created before cargo started adding one on crate creation :-)

Boddlnagg:
I thought about what you said about the approach similar to PCM, but came up with this minimal prototype, which shows that it won't solve the problem ([playpen](https://is.gd/GtIwP8)):

```rust
use std::marker::PhantomData;

struct Seq;

impl Seq {
    fn get_io<'a>(&'a self) -> IO<'a> {
        IO(self)
    }
}

struct IO<'a>(&'a Seq);

struct Event<'a>(PhantomData<&'a ()>);

impl<'a> IO<'a> {
    fn event_input<'b>(&'b mut self) -> Event<'b> {
        Event(PhantomData)
    }
}

fn main() {
   let s = Seq;
   let mut io = s.get_io();
   let ev1 = io.event_input();
   let mut io2 = s.get_io(); // if `got_io` would take &mut self instead, that wouldn't be allowed, but then we're back where we were
   let ev2 = io2.event_input();
   (ev1, ev2); // ev1 is still valid here, but shouldn't be!
}
```

diwic:
The PCM struct has an extra `Cell<bool>`, i e runtime verification that this is not allowed.

src/seq.rs
Boddlnagg: Yes, you're right.

Boddlnagg:
Oh, I see ... I should have had a closer look, sorry. That approach would work. Just needs a different name for `IO`, because it's only for input ... unfortunately I don't have a good idea.

Boddlnagg:
I will make the requested changes tomorrow, when I'm back on a Linux machine. Using Windows here ;-)

src/seq.rs
diwic: For me it's the other way around, like "`Option<Cow<'a, [u8]>>`, I know what that is", but for EventExtraData it's "What's that for, gotta go look it up". So I'd prefer Option<Cow<'a, [u8]>>.

src/seq.rs
Boddlnagg: Also, my solution did not introduce a way to retrieve the referenced buffer without copying (i.e., a version of `get_data` with `Event<'a> -> &'a [u8]`. I think, there should be a special cased version for that as well (`get_data_borrowed`, to keep the names consistent?).

src/seq.rs
Boddlnagg: I did it in `new` because the length flags only depend on the event type, and `Event` has no method to change the type, so it's not necessary to "fix" it whenever `ensure_buf` is called. Events that are returned from ALSA should always have the correct flags set, so it's only a problem when they are created manually using `Event::new`.

src/seq.rs
Boddlnagg: I agree that it should return `Event<'static>`, but the name `into_static` has no precedence in the standard library or the [Rust naming conventions](https://aturon.github.io/style/naming.html). I'm not totally happey with `into_owned`, either, but I'd prefer it over `into_static`.

src/seq.rs
Boddlnagg: Yet another comment about this: My changes allowed creation of a sysex event from a `Vec<u8>` by move, so the `Vec` doesn't need to be copied. This is not possible with the original design, because `Event::new` takes `&D`, so it only has a reference of the vector and needs to copy it. I think the best way forward here is the following: 
Create a constructor especially for event types with variable length data (e.g. `Event::new_varlen`) that takes a `T: Into<Cow<'a, [u8]>>`. That way you can pass either a slice or a vector (by value).
A corresponding `get_data_varlen` should return `&'a [u8]`. What do you think?

diwic: > Create a constructor especially for event types with variable length data (e.g. Event::new_varlen) that takes a T: Into<Cow<'a, [u8]>>. That way you can pass either a slice or a vector (by value).

Sounds good to me. As for the naming, I think `new_ext` and `get_ext` could be a good option, as it goes with the `snd_seq_ev_ext_t` type.

 > A corresponding get_data_varlen should return &'a [u8]. What do you think?

`Option<&'a [u8]>` I assume. Do we need a version where you can move the Vec out (called `take_ext` or `into_ext`) without cloning?

:+1: otherwise

Boddlnagg: I don't think that we need `take_ext` ... at least I can't think of any usecase: If you're creating the Event yourself, you usually don't need to get the data out again (because you have put it there in the first place), and if the Event is obtained from an ALSA call, it won't have a `Vec` stored inside.

Boddlnagg: I think I should also remove the implementation of `EventData` for `Vec<u8>`. The documentation should refer the user to `new_ext` and `get_ext` for event types that carry variable-length data. There shouldn't be two ways to do the same thing, especially when one is less efficient than the other.

Boddlnagg:
I made the changes, including an `Input` struct for the methods that can mess with the input buffer.

diwic:
I looked it through and felt that it all looked good. Perhaps a comment explaining "Input" and why we would need it would be nice.

Boddlnagg:
Thanks for merging! I added such a comment internally for `event_input`, but I guess you're referring to a doc-comment directly on `Input`?

diwic:
Perhaps something on `seq::input()` would be nice, explaining why you can't do it twice.

PR #25, opened 2017-04-03
Add .gitignore

Boddlnagg:
This is the default one that Cargo creates nowadays.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+target/
+**/*.rs.bk
+Cargo.lock


.gitignore
```diff
@@ -0,0 +1,3 @@
+target/
+**/*.rs.bk
```
diwic: What is this one?

Boddlnagg: I don't know for sure, I guess it's about backup files that some editors might create. This line is part of the `.gitignore` that `cargo new` creates, so I included it. I can remove it, if you prefer.

diwic: Weird. My `cargo new` does not create such a file.

PR #35, opened 2018-05-10
Silence note on Noteon event with 0 velocity.

jkristell:
My Miditech MIDI controller sends Noteon events with the velocity set to
0 instead of sending the Noteoff event. This change makes the synth
behave as expected.


diwic:
Good catch! Thanks for the patch.

PR #37, opened 2018-09-16
added missing mapping for few functions being used by arecord

vim-zz:


vim-zz:
Hi, thanks, it would be great

PR #39, opened 2018-11-15
Readability improvements to synth example

WaDelma:
Some clean-up to make the example more readable.


diwic:
What is readable and not is much a matter of taste, but I guess your taste is as good as mine, so...merged. :-)

PR #40, opened 2018-11-17
Add a constructor with &str argument for HintIter

mpevnev:
Just a convenience thing. I hope temporarily creating a `CString` (and unwrapping it) just for a single function call is not against the lib's philosophy. 

I'm not particularly happy with the function's name, any suggestions?

### DIFF
--- a/src/device_name.rs
+++ b/src/device_name.rs
@@ -36,6 +36,11 @@ impl HintIter {
         acheck!(snd_device_name_hint(cnr, iface.as_ptr(), &mut p))
             .map(|_| HintIter(p, 0))
     }
+
+    /// A constructor variant that takes the interface as a Rust string slice.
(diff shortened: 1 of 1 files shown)


diwic:
> Just a convenience thing. I hope temporarily creating a CString (and unwrapping it) just for a single function call is not against the lib's philosophy.

It is fine. It's tricky though, because in theory alsa naming is not guaranteed to be UTF-8, although in practice I've never seen anything above 127 at all.

 > I'm not particularly happy with the function's name, any suggestions?

Hmm...I guess it should be short if the main reason for this convenience function is saving characters at the caller's site... maybe just `new_str` or something.

mpevnev:
I've renamed the function, hopefully it's still clear what it does.

diwic:
Merged now, let me know if you need a release. Thanks!

PR #45, opened 2019-07-26
Add methods

derekdreery:
Adds a selection of alsa methods (mostly HwParams methods)

### DIFF
--- a/src/pcm.rs
+++ b/src/pcm.rs
@@ -570,11 +570,17 @@ impl<'a> HwParams<'a> {
 
     pub fn get_rate_max(&self) -> Result<u32> {
         let mut v = 0;
+        // Note on the null ptr: if this ptr is not null, then the value behind it is replaced with
+        // -1 if the suprenum is not in the set (i.e. it's an open range), 0 otherwise. This could
+        // be returned along with the value, but it's safe to pass a null ptr in, in which case the
+        // pointer is not dereferenced.
         acheck!(snd_pcm_hw_params_get_rate_max(self.0, &mut v, ptr::null_mut())).map(|_| v as u32)
     }
 
     pub fn get_rate_min(&self) -> Result<u32> {
         let mut v = 0;
+        // Note on the null ptr: see get_rate_max but read +1 and infinum instead of -1 and
+        // suprenum.
         acheck!(snd_pcm_hw_params_get_rate_min(self.0, &mut v, ptr::null_mut())).map(|_| v as u32)
     }
 
@@ -592,6 +598,8 @@ impl<'a> HwParams<'a> {
             .and_then(|_| Format::from_c_int(v, "snd_pcm_hw_params_get_format"))
     }
 
+    /// Returns true if the supplied format is valid for the pcm device this HwParams is attached
+    /// to.
     pub fn test_format(&self, v: Format) -> bool {
         unsafe { alsa::snd_pcm_hw_params_test_format((self.1).0, self.0, v as c_int) == 0 }
     }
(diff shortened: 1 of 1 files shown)


src/pcm.rs
```diff
@@ -544,6 +568,20 @@ impl<'a> HwParams<'a> {
         acheck!(snd_pcm_hw_params_get_rate(self.0, &mut v, &mut d)).map(|_| v as u32)
     }
 
+    pub fn get_rate_max(&self) -> Result<u32> {
+        let mut v = 0;
+        acheck!(snd_pcm_hw_params_get_rate_max(self.0, &mut v, ptr::null_mut())).map(|_| v as u32)
```
diwic: The `null_mut` looks weird here. I'm not saying it's wrong, but could you explain why you think this is the right thing to do?

src/pcm.rs
```diff
@@ -554,6 +592,10 @@ impl<'a> HwParams<'a> {
             .and_then(|_| Format::from_c_int(v, "snd_pcm_hw_params_get_format"))
     }
 
+    pub fn test_format(&self, v: Format) -> bool {
```
diwic: Please add a short comment like "returns true if the format is available" (just because it differs from what alsa-lib does)

src/pcm.rs
```diff
@@ -619,6 +671,20 @@ impl<'a> HwParams<'a> {
         acheck!(snd_pcm_hw_params_get_buffer_time_max(self.0, &mut v, &mut d)).map(|_| v as u32)
     }
 
+    /// This function should only be called when the configuration space contains a single
```
diwic: It seems a bit weird to have copied this additional comment but not the description of the function (i e "Check if hardware supports pause.").

src/pcm.rs
```diff
@@ -639,9 +705,14 @@ impl<'a> Clone for HwParams<'a> {
 
 impl<'a> fmt::Debug for HwParams<'a> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f,
-           "HwParams(channels: {:?}, rate: {:?} Hz, format: {:?}, access: {:?}, period size: {:?} frames, buffer size: {:?} frames)",
-           self.get_channels(), self.get_rate(), self.get_format(), self.get_access(), self.get_period_size(), self.get_buffer_size())
+        f.debug_struct("HwParams")
+            .field("channels", &self.get_channels())
+            .field("rate", &format!("{:?}Hz", self.get_rate()))
```
diwic: Missing space before "Hz"

src/pcm.rs
derekdreery: Ok after some extensive digging in the alsa source, I can explain this.

The param is actually a return type. If a non-null pointer is supplied, the pointer is dereferenced and the value at that location is updated to inform the caller whether the maximum value is actually included. So basically if there are a range of allowed rates, then after calling snd_pcm_hw_params_get_rate_max the value behind the ptr will be -1 if the set is open (indicating the actual max is less than the returned value), or 0 if closed (indicating that the returned value is included in the interval).

If you pass a null pointer (i.e. 0), then it is ignored: all modifications to the value behind the pointer are preceded with `if (ptr)`.

The documentation is awful for alsa, I will write something in the docs in this PR about all this.

derekdreery: I've added some notes about this to the source. They are maths-y but I think that's ok since they won't appear in the docs, they are just there to justify the use of null ptr.

derekdreery:
Thanks for making a great lib and being super responsive btw :)

diwic:
Sorry, I totally forgot this PR, so much for being "super responsive"...

Merged now, although I changed the `test_`-functions to return a `Result`, as that made more sense to me on second thought.

derekdreery:
hehe no worries!

PR #46, opened 2019-10-26
Feature/seq additions

flowkclav:
Had to add support for a few additional ALSA functions for a project I'm working on, thought I would contribute.

### DIFF
--- a/src/seq.rs
+++ b/src/seq.rs
@@ -193,7 +193,7 @@ impl Seq {
         Input::new(self)
     }
     
-    pub fn remove_events(&self, condition: RemoveEventsInfo) -> Result<()> {
+    pub fn remove_events(&self, condition: RemoveEvents) -> Result<()> {
         acheck!(snd_seq_remove_events(self.0, condition.0)).map(|_| ())
     }    
 }
@@ -476,7 +476,7 @@ bitflags! {
 
 bitflags! {
     /// [SND_SEQ_REMOVE_xxx]https://www.alsa-project.org/alsa-doc/alsa-lib/group___seq_event.html) constants 
-    pub struct RemoveEventsCondition: u32 {
+    pub struct Remove: u32 {
         const REMOVE_INPUT = (1<<0);
         const REMOVE_OUTPUT = (1<<1);
         const REMOVE_DEST = (1<<2);
@@ -560,25 +560,25 @@ impl PortSubscribe {
 
 /// [snd_seq_query_subs_type_t](https://www.alsa-project.org/alsa-doc/alsa-lib/group___seq_subscribe.html) wrapper
 #[derive(Copy, Clone)]
-pub enum SubscriptionQueryType {
+pub enum QuerySubsType {
     READ = alsa::SND_SEQ_QUERY_SUBS_READ as isize,
     WRITE = alsa::SND_SEQ_QUERY_SUBS_WRITE as isize,
 }
 
 /// [snd_seq_query_subscribe_t](https://www.alsa-project.org/alsa-doc/alsa-lib/group___seq_subscribe.html) wrapper
 //(kept private, functionality exposed by PortSubscribeIter)
-struct SubscriptionQueryInfo(*mut alsa::snd_seq_query_subscribe_t);
+struct QuerySubscribe(*mut alsa::snd_seq_query_subscribe_t);
 
-unsafe impl Send for SubscriptionQueryInfo {}
+unsafe impl Send for QuerySubscribe {}
 
-impl Drop for SubscriptionQueryInfo {
+impl Drop for QuerySubscribe {
     fn drop(&mut self) { unsafe { alsa::snd_seq_query_subscribe_free(self.0) } }
 }
 
-impl SubscriptionQueryInfo {
+impl QuerySubscribe {
     pub fn new() -> Result<Self> {
         let mut q = ptr::null_mut();
-        acheck!(snd_seq_query_subscribe_malloc(&mut q)).map(|_| SubscriptionQueryInfo(q))
+        acheck!(snd_seq_query_subscribe_malloc(&mut q)).map(|_| QuerySubscribe(q))
     }
 
     pub fn get_index(&self) -> i32 { unsafe { alsa::snd_seq_query_subscribe_get_index(self.0) as i32 } }
@@ -595,7 +595,7 @@ impl SubscriptionQueryInfo {
         let a = alsa::snd_seq_addr_t { client: value.client as c_uchar, port: value.port as c_uchar};
         alsa::snd_seq_query_subscribe_set_root(self.0, &a);
     } }
-    pub fn set_type(&self, value: SubscriptionQueryType) { unsafe {
+    pub fn set_type(&self, value: QuerySubsType) { unsafe {
         alsa::snd_seq_query_subscribe_set_type(self.0, value as alsa::snd_seq_query_subs_type_t)
     } }
     pub fn set_index(&self, value: i32) { unsafe { alsa::snd_seq_query_subscribe_set_index(self.0, value as c_int) } }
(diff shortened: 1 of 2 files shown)


diwic:
Hi and thanks for the PR!

As a general comment, I'd like the naming convention to wrap alsa-lib's a little closer, e g, `SND_SEQ_REMOVE_xxx` should be named `Remove` (since we're in the `seq` module) and have constants named `INPUT`, `OUTPUT` etc. `reset_all` should be named `init`. And so on. Even if alsa-lib's naming is not always the best, there is value in being able to translate C examples straight over to Rust, look up documentation etc.
 
Is this something you'd be willing to change? (If not, I might take the PR as it is and change it myself.)

flowkclav:
Hello!

For the `SND_SEQ_REMOVE_xxx` stuff, I realised I was diverging from your established convention.  The reason I went ahead is that the `bitflag!` macro puts all those names in the module's namespace and I didn't want to pollute the namespace with generic names such as `INPUT` or `DEST`. (I don't think any of them would conflict but my thought was it might be confusing along side alsa::seq::`READ` and others).   That being said, if you still prefer I remove the `REMOVE_` prefix from these flags, I have no problem doing so.

For `reset_all`, I just got inspired by the existing `enable_running_status` and the ALSA doc for the `init` function that mentions "resets MIDI encode/decode parsers" (for which I found `init` to be a bad name).   I went back and forth on that one before committing, have no problem with calling it `init`.

In both cases I don't mind pushing another commit with those changes, let me know.

src/seq.rs
```diff
@@ -457,6 +474,22 @@ bitflags! {
     }
 }
 
+bitflags! {
+    /// [SND_SEQ_REMOVE_xxx]https://www.alsa-project.org/alsa-doc/alsa-lib/group___seq_event.html) constants 
+    pub struct RemoveEventsCondition: u32 {
```
diwic: Rename to "Remove"

src/seq.rs
```diff
@@ -525,6 +558,94 @@ impl PortSubscribe {
 
 }
 
+/// [snd_seq_query_subs_type_t](https://www.alsa-project.org/alsa-doc/alsa-lib/group___seq_subscribe.html) wrapper
+#[derive(Copy, Clone)]
+pub enum SubscriptionQueryType {
```
diwic: Rename to "QuerySubsType"

src/seq.rs
```diff
@@ -525,6 +558,94 @@ impl PortSubscribe {
...
+    READ = alsa::SND_SEQ_QUERY_SUBS_READ as isize,
+    WRITE = alsa::SND_SEQ_QUERY_SUBS_WRITE as isize,
+}
+
+/// [snd_seq_query_subscribe_t](https://www.alsa-project.org/alsa-doc/alsa-lib/group___seq_subscribe.html) wrapper
+//(kept private, functionality exposed by PortSubscribeIter)
+struct SubscriptionQueryInfo(*mut alsa::snd_seq_query_subscribe_t);
```
diwic: Rename to "QuerySubscribe"

src/seq.rs
```diff
@@ -525,6 +558,94 @@ impl PortSubscribe {
...
+    } }
+    pub fn set_type(&self, value: SubscriptionQueryType) { unsafe {
+        alsa::snd_seq_query_subscribe_set_type(self.0, value as alsa::snd_seq_query_subs_type_t)
+    } }
+    pub fn set_index(&self, value: i32) { unsafe { alsa::snd_seq_query_subscribe_set_index(self.0, value as c_int) } }
+}
+
+#[derive(Copy, Clone)]
+/// Iterates over port subscriptions for a givent client:port/type.
+pub struct PortSubscribeIter<'a>(&'a Seq, Addr, SubscriptionQueryType, i32);
```
diwic: If we have more than two - three fields, it's better to make a struct with named fields.

src/seq.rs
```diff
@@ -1119,6 +1240,111 @@ impl QueueTempo {
...
+    pub fn get_tick_time(&self) -> u32 { unsafe {alsa::snd_seq_queue_status_get_tick_time(self.0) as u32 } }
+    pub fn get_real_time(&self) -> time::Duration { unsafe {
+        let t = &(*alsa::snd_seq_queue_status_get_real_time(self.0));
+        time::Duration::new(t.tv_sec as u64, t.tv_nsec as u32)
+    } }
+    pub fn get_status(&self) -> u32 { unsafe { alsa::snd_seq_queue_status_get_status(self.0) as u32 } }
+}
+
+/// [snd_seq_remove_events_t](https://www.alsa-project.org/alsa-doc/alsa-lib/group___seq_event.html) wrapper
+pub struct RemoveEventsInfo(*mut alsa::snd_seq_remove_events_t);
```
diwic: Rename to "RemoveEvents"

src/seq.rs
```diff
@@ -1119,6 +1240,111 @@ impl QueueTempo {
...
+    pub fn get_queue(&self) -> i32 { unsafe { alsa::snd_seq_remove_events_get_queue(self.0) as i32 } }
+    pub fn get_time(&self) -> time::Duration { unsafe {
+        let mut d = alsa::snd_seq_timestamp_t { data: (*alsa::snd_seq_remove_events_get_time(self.0)).data };
+        let t = &(*d.time());
+        
+        time::Duration::new(t.tv_sec as u64, t.tv_nsec as u32)
+    } }
+    pub fn get_tick(&self) -> u32 { unsafe {
+        let mut d = alsa::snd_seq_timestamp_t { data: (*alsa::snd_seq_remove_events_get_time(self.0)).data };
+
```
diwic: `alsa::snd_seq_timestamp_t` is a union so it contains either a tick or a time, or both. 
Can you double check that both `get_time` and `get_tick` be used the way you do here?

src/seq.rs
```diff
@@ -1119,6 +1240,111 @@ impl QueueTempo {
...
+        let mut t = &mut (*d.time());
+        
+        t.tv_sec = value.as_secs() as c_uint;
+        t.tv_nsec = value.subsec_nanos() as c_uint;
+
+        alsa::snd_seq_remove_events_set_time(self.0, &d);
+    } }
+    pub fn set_tick(&self, value: u32) { unsafe {
+        let mut d = alsa::snd_seq_timestamp_t {data: [0; 2]};
+
```
diwic: Same concern as for get_time and get_tick apply to set_time and set_tick.

src/seq.rs
```diff
@@ -1141,6 +1367,13 @@ impl MidiEvent {
     /// Alsa-lib is a bit confusing here. Anyhow, set "enable" to true to enable running status.
     pub fn enable_running_status(&self, enable: bool) { unsafe { alsa::snd_midi_event_no_status(self.0, if enable {0} else {1}) } }
 
+    /// Resets both encoder and decoder
+    pub fn reset_all(&self) { unsafe { alsa::snd_midi_event_init(self.0) } }
```
diwic: Let's rename it to "init"

src/seq.rs
diwic: Also I'm not sure if there is a reason not to make it public?

diwic:
> the bitflag! macro puts all those names in the module's namespace and I didn't want to pollute the namespace with generic names such as INPUT or DEST.

Are you sure? A quick look at the `bitflags` crate, from the first example it looks like the namespace is at the struct level, rather than at the module level.

src/seq.rs
flowkclav: Renamed.  (Not sure what etiquette is for these change requests... I'll let you resolve the conversations)

src/seq.rs
flowkclav: Renamed.

src/seq.rs
flowkclav: Renamed.

The reason I kept it private is because this struct is used internally by PortSubscribeIter, but would have no real purpose to users of this crate (they could create it and get/set values but would do not useful work).  That being said if you prefer it public let me know, I'll change it.

src/seq.rs
flowkclav: Changed. (Yes, wasn't sure about that one. A lot cleaner with named fields. I kept the fields private since they were private when it was a tuple)

src/seq.rs
flowkclav: Renamed.

src/seq.rs
flowkclav: Oups, my bad. good catch.   Definitely used as time here, not tick. Removed the tick getters/setters.

src/seq.rs
flowkclav: Removed.

src/seq.rs
flowkclav: Renamed.

flowkclav:
> > the bitflag! macro puts all those names in the module's namespace and I didn't want to pollute the namespace with generic names such as INPUT or DEST.
> 
> Are you sure? A quick look at the `bitflags` crate, from the first example it looks like the namespace is at the struct level, rather than at the module level.

It looks like it is in the module's namespace though (unless I'm completely out of left field and we're talking about two different things :)).  In the crate's doc's sample,  the flags defined as

```
bitflags! {
    struct Flags: u32 {
        const FLAG_A       = 0b00000001;
...
} }
```
are accessed with `let e1 = FLAG_A | FLAG_C;` and not `let e1 = Flags::FLAG_A | Flags::FLAG_C;`

Similarly, in my application I refer to the remove flags directly with e.g. `alsa::seq::REMOVE_OUTPUT` and not `alsa::seq::Remove::REMOVE_OUTPUT`.

That was my justification for adding "REMOVE_", but at the end of the day if you want it gone I can make the change.

diwic:
Thanks for being helpful! Sounds like I should have a look at upgrading the bitflags dependency.

PR #47, opened 2019-10-29
Add period and buffer size min/max getters on HwParams

fredszaq:
These can be quite usefull when setuping a device


diwic:
The change to gitignore is unrelated but I decided to let it slip through. Thanks for the PR!

PR #48, opened 2020-01-09
add handle_events to mixer api

okeri:
I'm not sure about catching events api,
but anyway i guess functions like snd_mixer_handle_events and/or snd_hctl_handle_events should be wrapped even using poll.


diwic:
Thanks!

ashthespy:
Just wondering - would you have an example of how to use this?

PR #49, opened 2020-01-10
Unbreak build on BSDs

jbeich:
`nfds_t` is `unsigned int` on Android, DragonFly, FreeBSD, NetBSD all of which can use ALSA. See https://github.com/rust-lang/libc/search?q=nfds_t


diwic:
Thanks!

PR #51, opened 2020-03-19
rawmidi: Fix non-blocking open on input devices

chrissie-c:
Attempting to use non-blocking IO on an input device in rawmidi causes an assert in libasound:

midirust: rawmidi_hw.c:213: snd_rawmidi_hw_open: Assertion `outputp' failed.

This seems to be because we are passing SND_RAWMIDI_APPEND (1) instead of SND_RAWMIDI_NONBLOCK (2) into rawmidi_open and there is no output device.

This patch would be better if it had the proper constants in, of course, but this fixes it for the time-being.

### DIFF
--- a/src/rawmidi.rs
+++ b/src/rawmidi.rs
@@ -119,7 +119,7 @@ impl Rawmidi {
 
     pub fn open(name: &CStr, dir: Direction, nonblock: bool) -> Result<Rawmidi> {
         let mut h = ptr::null_mut();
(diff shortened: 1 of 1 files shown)


diwic:
Thanks!

diwic:
Let me know if you need a release with this patch in.

chrissie-c:
Thanks for pulling. I'm not in a rush for a release, this is just a home project so I can live with a private build until you're ready.

PR #54, opened 2020-04-27
Expose `snd_pcm_delay`

mitchmindtree:
This is useful for determining the instant at which the currently
buffered samples will have completed playback.


diwic:
Thanks! (Although you can use `avail_delay` to get the same data already.)

PR #55, opened 2020-05-16
Re-export libc::pollfd

F35idk:
So library users don't have to put libc in their Cargo.toml in order to for example store pollfd's in a struct (or otherwise pass around pollfd's explicitly).


diwic:
Thanks!

PR #57, opened 2020-11-05
Helpers for seq::Addr

calvinkosmatka:
Hi @diwic, wanted to add a couple of helpers for working with sequencer addresses. One to get an Addr from a PortInfo struct, and a FromStr implementation for addresses in the form "[client]:[port]". Let me know what you think!

### DIFF
--- a/src/seq.rs
+++ b/src/seq.rs
@@ -5,6 +5,7 @@ use super::error::*;
 use crate::alsa;
 use super::{Direction, poll};
 use std::{ptr, fmt, mem, slice, time, cell};
+use std::str::{FromStr, Split};
 use std::ffi::{CStr};
 use std::borrow::Cow;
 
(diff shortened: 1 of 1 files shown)


src/seq.rs
```diff
@@ -369,6 +371,13 @@ impl PortInfo {
         unsafe { alsa::snd_seq_port_info_set_type(self.0, c.bits() as c_uint) }
     }
 
+    pub fn address(&self) -> Addr {
```
diwic, line 374: Shorten address to "addr" and add a comment - functions that do not represent a 1:1 relationship with an alsa-lib function should have a comment.

src/seq.rs
```diff
@@ -474,6 +483,19 @@ pub struct Addr {
...
 }
 
+impl FromStr for Addr {
+    type Err = ParseIntError;
+
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        let split: Vec<&str> = s.trim()
+                                .split(":")
+                                .collect();
+        let client = split[0].parse::<i32>()?;
```
diwic, line 493: You don't need to collect into a Vec - just call `next()` on the iterator itself.

Also, it does not error out correctly in case you have too many or too few ':', i e, it should not panic on strings like '', '123', and '1:2:3'.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "alsa"
-version = "0.4.2"
+version = "0.4.3"
```
diwic, line 3: Let's skip the version changing commit for this PR - I can do a release after this PR if you like?

diwic (changes requested):
Hi and thanks for the PR!

I think these are good changes in general and will be happy to merge it once you have addressed the review comments.

src/seq.rs
calvinkosmatka: Updated in the latest commit

Cargo.toml
calvinkosmatka: That makes sense to me.

src/seq.rs
calvinkosmatka: Shortened and commented!

diwic:
Thanks, looks good now. Would you mind squashing it into one patch before I merge it?

calvinkosmatka:
Done. Thanks for reviewing!

diwic:
Thanks for the contribution!

PR #60, opened 2020-12-11
Update dependencies. 

hrydgard:
Also makes synth-example use the local alsa-rs, so you can easily test changes.

As for the error change, .description() is deprecated on Error.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ readme = "README.md"
 edition = "2018"
 
 [dependencies]
-libc = "0.2.81"
+libc = "0.2.65"
 alsa-sys = "0.3.0"
 bitflags = "1.2.1"
 nix = "0.19.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,10 +13,10 @@ readme = "README.md"
 edition = "2018"
 
 [dependencies]
-libc = "0.2.65"
+libc = "0.2.81"
 alsa-sys = "0.3.0"
 bitflags = "1.2.1"
-nix = "0.15"
+nix = "0.19.1"
 
 [badges]
 is-it-maintained-issue-resolution = { repository = "diwic/alsa-rs" }
--- a/synth-example/Cargo.toml
+++ b/synth-example/Cargo.toml
@@ -5,5 +5,5 @@ authors = ["David Henningsson <diwic@ubuntu.com>"]
 
 [dependencies]
 sample = "0.7"
-alsa = "0.4"
+alsa = { path = ".." }
 clap = "2.20"
--- a/synth-example/Cargo.toml
+++ b/synth-example/Cargo.toml
@@ -5,5 +5,5 @@ authors = ["David Henningsson <diwic@ubuntu.com>"]
 
 [dependencies]
 sample = "0.7"
-alsa = { path = ".." }
+alsa = { path = "..", version = "0.4" }
 clap = "2.20"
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -13,10 +13,10 @@ readme = "README.md"
 edition = "2018"
 
 [dependencies]
-libc = "0.2.65"
+libc = "0.2.81"
 alsa-sys = "0.3.0"
```
diwic, line 17: Is there anything bad with existing versions of libc/nix that motivates the dependency update? AFAIK, "0.2.65" means  "0.2.65 or higher".

synth-example/Cargo.toml
```diff
@@ -5,5 +5,5 @@ authors = ["David Henningsson <diwic@ubuntu.com>"]
 
 [dependencies]
 sample = "0.7"
-alsa = "0.4"
+alsa = { path = ".." }
 clap = "2.20"
```
diwic, line 9: This should actually be 
`alsa = { path = "..", version = "0.4" }`

synth-example/src/main.rs
```diff
@@ -312,5 +312,5 @@ fn run() -> Res<()> {
 }
 
 fn main() {
-    if let Err(e) = run() { println!("Error ({}) {}", e.description(), e); }
+    if let Err(e) = run() { println!("Error ({})", e); }
 }
```
diwic, line 316: This one's ok

diwic (changes requested):
Hi and thanks for the PR! Se review comments.

Cargo.toml
hrydgard: Nah, doesn't matter. I'll revert this part.

diwic:
@hrydgard You forgot to revert the nix dependency update (and please squash the patch when you think it's ready for merge), but that does not explain why you wanted to do it in the first place, and also I'm curious how this PR helps Embark studios in making alsa-rs more game friendly?

hrydgard:
Oh I very intentionally left the nix one since it's a "major" version upgrade that won't be auto-reconciled, unlike the libc one. 

Just doing some duplicate dependency cleanup. We ended up with two versions of nix in our build - which is normal, but trying to keep the number of dupes in check (we use cargo-deny to find them). Actually the other version we pull in through another dependency is 0.18 so this change won't make things strictly better until I can update those other dependencies too.

Boddlnagg:
It's also a bit problematic that alsa-rs makes `nix` part of its public API without reexporting it (e.g. in [`errno`](https://docs.rs/alsa/0.4.3/alsa/struct.Error.html#method.errno)). This means that downstream crates need to always use a version that is compatible with the one used by alsa-rs when they want to use these parts of the API. When multiple dependencies would do that, and they use mutually incompatible versions, that's a problem (although I haven't run into it yet).

See also https://github.com/Boddlnagg/midir/pull/66#issuecomment-670842508.

diwic:
@Boddlnagg I believe this is now fixed in https://github.com/diwic/alsa-rs/commit/78787a911fdcfb8880bfef30f24e9665e7d1ef81 - can you verify?

Boddlnagg:
I can't at the moment unfortunately, because I have no access to a Linux environment right now. But looking at the code, it seems correct. You can verify yourself by checking if you can in a downstream crate do something like `e.errno() == Some(nix::errno::Errno::ENOSPC)` without depending on `nix` directly.

PR #61, opened 2020-12-15
Add `AudioTstampType` and related funtions.

alexmoon:


src/pcm.rs
```diff
@@ -923,6 +949,16 @@ impl Status {
     }
 }
 
+/// The type of timestamp returned by `Status::get_audio_htstamp()`.
+/// These values are derived from the [audio_time](https://github.com/alsa-project/alsa-lib/blob/master/test/audio_time.c)
+/// test program in the ALSA library.
+pub enum AudioTstampType {
```
diwic, line 955: Hmm, but looking at the referenced file above, there are six different values (lines 272-283), not four. Could you explain?

src/pcm.rs
```diff
@@ -923,6 +949,16 @@ impl Status {
...
+    Default = 0,
+    Link = 1,
+    LinkEstimated = 2,
+    Synchronized = 3,
+}
```
diwic, line 960: Probably this enum should be non-exhaustive

src/pcm.rs
```diff
@@ -178,6 +178,23 @@ impl PCM {
...
+    ) -> Result<Status> {
+        let z = Status::new();
+        let mut cfg = alsa::snd_pcm_audio_tstamp_config_t {
+            _bitfield_1: alsa::snd_pcm_audio_tstamp_config_t::new_bitfield_1(
+                type_requested as libc::c_uint,
+                libc::c_uint::from(report_delay),
+            ),
+            __bindgen_padding_0: [0; 3],
+        };
+        unsafe { alsa::snd_pcm_status_set_audio_htstamp_config(z.ptr(), &mut cfg) };
```
diwic, line 194: I would prefer if we avoid the names `_bitfield_1` and `__bindgen_padding_0` when possible. Bindgen has a tendency to change them occasionally and it's good to be resilient (also IMO the code will be more readable).
I e, 

    let mut cfg = mem::zeroed();
    cfg.set_requested_type(...);
    cfg.set_report_delay(...);

src/pcm.rs
```diff
@@ -178,6 +178,23 @@ impl PCM {
         acheck!(snd_pcm_status(self.0, z.ptr())).map(|_| z)
     }
 
+    pub fn status_with_audio_htstamp_config(
+        &self,
```
diwic, line 182: Functions that do not map 1-to-1 to an alsa-lib function should have a comment explaining what they do.

diwic (changes requested):
Hi and thanks for the PR! Please see the review comments.

src/pcm.rs
alexmoon: Good catch. Unfortunately I haven't been able to find any decent documentation for the values. I was going off the usage string, I'd expect live code to be more accurate. I'll change the enum to use those values.

src/pcm.rs
alexmoon: Agreed. I wasn't sure if there was a MSRV issue so I left it off, but I'll add it.

src/pcm.rs
alexmoon: Yes, that's better. I haven't worked much with bindgen. Thanks for the suggestion.

src/pcm.rs
alexmoon: Will do. Thanks for the helpful feedback.

diwic:
@alexmoon So I did some digging and found the constants and figured that the right thing to do was to send a patch upstream to alsa-lib. If they do we can then regenerate through alsa-sys and get the constants that way.

As for the `#[non_exhaustive]` part that should probably be added to more enums (now that I think of it),  but I guess that attribute was not in Rust when I started writing on those bindings.

alexmoon:
> @alexmoon So I did some digging and found the constants and figured that the right thing to do was to send a patch upstream to alsa-lib. If they do we can then regenerate through alsa-sys and get the constants that way.

That makes sense to me.

alexmoon:
BTW, one other API I considered was instead of adding `PCM::status_with_audio_htstamp_config()` to create a `StatusBuilder` struct with a `set_audio_htstamp_config(..)` method and a `build(self, pcm: &PCM) -> Status` method. That would be more convenient if they add more settings in the future, but seemed like overkill for right now.

alexmoon:
So I thought about it some more and decided that the builder API was closer to the underlying ALSA API and in some ways conceptually simpler so I've switched to it. Once alsa-sys is updated with the new version of alsa-lib we should be able to wrap this PR up.

src/pcm.rs
```diff
@@ -949,6 +931,35 @@ impl Status {
...
+/// status from the stream.
+pub struct StatusBuilder(Status);
+
+impl StatusBuilder {
+    pub fn new() -> Self {
+        StatusBuilder(Status::new())
+    }
+
+    pub fn set_audio_htstamp_config(
+        self,
```
diwic, line 946: Builder pattern shouldn't have "set_" in its name. Probably just "audio_htstamp_config" is the right name here.

diwic:
@alexmoon so alsa-sys 0.3.1 has been released now with relevant upstream changes in it, you can update your PR. Please squash it when you think it's ready for merging.

alexmoon:
@diwic ok, I've squashed the commits. Let me know if there are any other changes you would like to see.

diwic:
Merged now. Thanks!

PR #65, opened 2021-03-10
update nix to 0.20

xMAC94x:
Hi,

could you please update the nix version used, and also `cargo publish` to crates.io, as the nix version there is quite old (0.15).
Thank you


diwic:
Okay, done now. Enjoy!

PR #66, opened 2021-04-15
Update Cargo.toml to match root package version

damirka:
I've tried to run synth example locally, and it failed to resolve dependency.

```bash
pi@raspberrypi:~/projects/alsa-rs/synth-example $ cargo run 
    Updating crates.io index
error: failed to select a version for the requirement `alsa = "^0.4"`
candidate versions found which didn't match: 0.5.0
location searched: /home/pi/projects/alsa-rs
required by package `synth-example v0.1.0 (~/projects/alsa-rs/synth-example)`
```

Fixing it to `0.5` solves the problem (and introduces new one 😆)

### DIFF
--- a/synth-example/Cargo.toml
+++ b/synth-example/Cargo.toml
@@ -5,4 +5,4 @@ authors = ["David Henningsson <diwic@ubuntu.com>"]
 
 [dependencies]
 sample = "0.7"
-alsa = { path = "..", version = "0.4" }
+alsa = { path = "..", version = "0.5" }


diwic:
> Fixing it to 0.5 solves the problem (and introduces new one)

Merged now, thanks. What new problem is introduced?

damirka:
Hey, @diwic. Thanks for merging.
Next one is that `sample` library can not be compiled (when I run `cargo run` over synth example).

And there's a lot happened to this library: https://github.com/RustAudio/dasp/pull/120, so it might be a good idea to rewrite this synth example.

diwic:
@damirka Right. Last time I checked both dasp and sample were broken, they stopped compiling due to a regression in the Rust compiler, but apparently dasp is fixed now and sample isn't, so I switched over now.

PR #72, opened 2021-06-26
Include license

remilauzier:


Cargo.toml
```diff
@@ -11,6 +11,7 @@ license = "Apache-2.0/MIT"
 categories = ["multimedia::audio", "api-bindings"]
 readme = "README.md"
 edition = "2018"
+include = ["README.md", "LICENSE-*", "Cargo.toml", "src/"]
```
diwic, line 14: I've just read up on "include" and "exclude" and I think this row is not needed. Just having the files in the repo should be enough to get them included...I think?

LICENSE-APACHE
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [2015-2021] [David Henningsson <diwic@ubuntu.com>]
```
diwic, line 190: Can we skip this section:

 1. I'm not the sole copyright holder of this repository - I've had contributors - so listing only me here seems unfair. 
 2. I also don't want to maintain a list of copyright holders manually. If this is needed you need to extract them from the git history during packaging.
 3. The email address is old and no longer works. I've just pushed an update to my email address.

LICENSE-MIT
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) [2015-2021] [David Henningsson <diwic@ubuntu.com>]
```
diwic, line 3: Same as previous comment

diwic (changes requested):

Cargo.toml
remilauzier: For some other package it was not working. Cargo can skip them if not include sometime.

LICENSE-APACHE
remilauzier: The license only need to have the creator and big major contributor names. From what i have see it's only you that need to be writen in this section.

PR #74, opened 2021-06-26
fix typo and fix order of impl

remilauzier:


src/seq.rs
```diff
@@ -1359,7 +1359,7 @@ impl MidiEvent {
     }
 
     /// In case of success, returns a tuple of (bytes consumed from buf, found Event).
-    pub fn encode<'a>(&'a mut self, buf: &[u8]) -> Result<(usize, Option<Event<'a>>)> {
+    pub fn encode(&mut self, buf: &[u8]) -> Result<(usize, Option<Event>)> {
         // The ALSA documentation clearly states that the event will be valid as long as the Encoder
```
diwic, line 1363: Because of the unsafe trickery going on inside this method, I'd like to keep this one as it is to make it clear what's going on.

src/seq.rs
```diff
@@ -915,10 +915,6 @@ impl EventData for EvNote {
              EventType::Noteoff |
              EventType::Keypress)
     }
-    fn get_data(ev: &Event) -> Self {
```
diwic, line 918: Perhaps better to switch the order of methods in the trait itself than to switch all the implementations?

diwic (changes requested):

src/seq.rs
remilauzier: ``pub fn encode(&'_ mut self, buf: &[u8]) -> Result<(usize, Option<Event<'_>>)> {``
Could it work that way?

diwic: That's still less clear than the original. Don't you agree?

remilauzier: Yeah.

src/seq.rs
diwic: I honestly still don't get this part. Why is the order of methods important? Are they supposed to be in alphabetic order? Or can they be in any order as long as they follow what the trait declaration does? Is there a coding convention document you're following, or is it just that you think it looks tidier that way?

remilauzier: The trait and the impl should have the same order but it can be what you want but the trait decide for the impl. Just say me what order you want i i will change it.

diwic:
@remilauzier Okay, merged now. In general, I'm happy for PR that fix typos. 
But PRs that change the coding style to someone's personal preference are less likely to get merged. 

If the coding style change is according to some official Rust coding style guide (e g clippy) I'm more likely to merge. (I don't care about clippy myself but it seems other people do.)

remilauzier:
You can configure rustfmt.toml and clippy.toml to have your preference established for your project. You can also do ``#![allow(clippy::something)]`` in the main.rs or lib.rs

diwic:
@remilauzier Fair enough, I've enabled some clippy lints now.
But when it comes to the order of traits and impls, neither clippy nor rustc warns about that. So who says they need to be in the same order?

remilauzier:
For that, it's just my IDE that recommend that for simplicity. IntelliJ IDEA. Thanks for the clippy lint that will help further contributor to know what you want.

PR #78, opened 2021-07-11
Add a "auto endianness" function for all listed formats with endianness variants.

JasonLG1979:

PR #79, opened 2021-07-12
Fix up for https://github.com/diwic/alsa-rs/pull/78

JasonLG1979:
Fix a typo and a couple accidental omissions.

### DIFF
--- a/src/pcm.rs
+++ b/src/pcm.rs
@@ -499,11 +499,17 @@ impl Format {
     #[cfg(target_endian = "little")] pub const fn u24_3() -> Format { Format::U243LE }
     #[cfg(target_endian = "big")] pub const fn u24_3() -> Format { Format::U243BE }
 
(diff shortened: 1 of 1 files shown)


JasonLG1979:
Oops...

diwic:
@JasonLG1979 
While I don't mind completeness of the implementation, be aware that many of these formats are extremely rare in practice. I e, the 18 and 20 bit variants, probably that's from some old sound card from the last century or so.

JasonLG1979:
> While I don't mind completeness of the implementation, be aware that many of these formats are extremely rare in practice. I e, the 18 and 20 bit variants, probably that's from some old sound card from the last century or so.

Oh, I know. In all likely I could have left it and no one would have noticed. The main reason I came here was to add `s24_3` because it actually is a  pretty widely used format for USB DAC's. The others were just to cover all the bases just because. They might never actually be used? Who knows?

PR #83, opened 2022-02-16
Fix example synth

Wint3rmute:
Hello!

I've noticed that after version `0.6` release the `synth-example` had `alsa-rs` pinned to version `0.5`. Fixed that along with a minor API change that broke the synth build.


diwic:
Thanks!

PR #84, opened 2022-05-20
Update nix to 0.24, limit features

rtzoeller:
This reduces compile times and removes memoffset as an indirect dependency.


rtzoeller:
This reduced the clean compile time from 3.9 seconds to 2.3 seconds on my machine.

PR #86, opened 2022-07-10
add handle_events to hctl api

peng1999:
This is similar to #48, adding snd_hctl_handle_events to Hctl to make Hctl::wait actually useful.

### DIFF
--- a/src/hctl.rs
+++ b/src/hctl.rs
@@ -66,6 +66,10 @@ impl HCtl {
         if p.is_null() { None } else { Some(Elem(self, p)) }
     }
 
+    pub fn handle_events(&self) -> Result<()> {
+        acheck!(snd_hctl_handle_events(self.0)).map(|_| ())
+    }
+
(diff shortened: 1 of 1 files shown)


diwic:
Looks good and mergable to me, any reason why this is a draft?

peng1999:
I was not sure about the return type of the function then. The documentation of [snd_hctl_handle_events](https://www.alsa-project.org/alsa-doc/alsa-lib/group___h_control.html#ga02d9dc1cdbdfe11880e3fe774f2034bd) says that it returns 0 or negative. But after reading [the source code](https://github.com/alsa-project/alsa-lib/blob/10e2490023e937e0a6a524c21f5f82176c94e010/src/control/hcontrol.c#L779) and doing some experiments, I found the function can also return positive values (at most time it is 1). So I changed the return type to `Result<u32>` like `Mixer::handle_events`.

I believe this PR is now ready to be merged.

diwic:
Ok, merged now. Thanks!

PR #89, opened 2022-10-20
bump to libc 0.2.135 and nix 0.25

mbodmer:


diwic:
Hi and thanks for the PR, could you motivate why it is needed?

mbodmer:
Well, it's not exactly needed, just keep going forward. I am trying to avoid double libc and nix dependencies, as CPAL also has bumped them. With libc it might be better to not specify the patch release and just keep it at "0.2" so it would leave the exact version to cargo's version resolution.

mbodmer:
I've just found that it's also documented that way in libc: https://crates.io/crates/libc

mbodmer:
I now also did the same for nix, this allows cargo to have more version combination. alsa-rs can handle them all, at least at this point in time

mbodmer:
Hope this helps, thanks for reviewing :-)

est31:
From cpal's point of view, it would be nice to have this merged and after that, a release would be really good. Thanks!

diwic:
@est31 @mbodmer Released v0.6.1 now, enjoy!

mbodmer:
Great, thanks.

PR #93, opened 2022-12-01
implement read/write from ElemValue

chadmed:
Necessary to actually get/set values for a given elem. Tested on amd64 and aarch64.

### DIFF
--- a/src/ctl_int.rs
+++ b/src/ctl_int.rs
@@ -65,6 +65,14 @@ impl Ctl {
             .map(|_| m as i64)
     }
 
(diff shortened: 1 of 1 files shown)


src/ctl_int.rs
```diff
@@ -65,6 +65,14 @@ impl Ctl {
             .map(|_| m as i64)
     }
 
+    pub fn elem_read(&self, val: &ElemValue) -> Result<i32> {
```
diwic, line 68: Must be `&mut ElemValue` because it changes `val`.

chadmed:
should be fixed now :)

diwic:
Thanks!

PR #96, opened 2023-01-07
Implement new methods for Selem and fix return value of Mixer::wait

mbyzhang:
* Implement `set_capture_volume_all`, `ask_playback_db_vol`, and `ask_capture_db_vol` for Selem.
* Fix return value os `Mixer::wait`. The return value of the inner `snd_mixer_wait` should be tested against zero (success) instead of one.

### DIFF
--- a/src/mixer.rs
+++ b/src/mixer.rs
@@ -61,8 +61,8 @@ impl Mixer {
         acheck!(snd_mixer_handle_events(self.0)).map(|x| x as u32)
     }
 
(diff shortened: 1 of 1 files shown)


diwic:
Thanks for the PR!

As for `snd_mixer_wait`, it seems to work differently from `snd_hctl_wait`, so probably it should return `Result<()>` instead of `Result<bool>`. The rest is fine.

mbyzhang:
Fixed.

Thanks!

PR #98, opened 2023-01-13
Add functions to pcm::Format to get various attributes

carlgreen:
I tried to do this in a similar way to existing code. I just did the functions I needed; if you would prefer to have all format functions from https://www.alsa-project.org/alsa-doc/alsa-lib/group___p_c_m___helpers.html implemented at once I can look at doing that.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -46,6 +46,11 @@ impl $name {
         Self::all().iter().find(|&&x| c == x as ::libc::c_int).map(|&x| x)
             .ok_or_else(|| Error::unsupported(s))
     }
+
+    #[allow(dead_code)]
+    fn to_c_int(&self) -> ::libc::c_int {
+        return *self as ::libc::c_int;
+    }
 }
 
 }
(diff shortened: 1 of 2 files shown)


diwic:
Hi and thanks for the PR!

Feel free to implement as many or as few functions as you see fit, but you should run the functions through `acheck` to handle errors better and remove the `get_` prefix from the function names (so they strictly match alsa-lib).

carlgreen:
All renamed, I think I copied that from somewhere else where the get was part of it. I've added `acheck` around most of the functions, except for `format_silence_16` which returns an unsigned int so can't return any errors (see https://www.alsa-project.org/alsa-doc/alsa-lib/group___p_c_m___helpers.html#ga5a52bb63323f463198dea3f3c6aca571). Because of that I also simplified its return from `Result<u16>` to just `u16`.

diwic:
Thanks, looks good now. Squash it into one commit and I'll merge it.

carlgreen:
hopefully i squashed this right? thanks.

diwic:
Yep, it's good, merged now. Let me know if you need a release.

carlgreen:
thanks, no need for an immediate release for me.

PR #99, opened 2023-03-30
Update nix to 0.26

glandium:


diwic:
For what reason? Usually updates cause more version churn than they're worth.

glandium:
The main reason would be that many crates have switched to nix 0.26 (the most used of which is tokio), and this would avoid depending on two versions of the crate. Alternatively, the dependency could be turned into >=0.24,<0.27

diwic:
@RustAudio/cpal-maintainers have you fixed up your stuff so I can do this without breaking cpal?

PR #101, opened 2023-04-30
implement FromStr for Format

### LINKED ISSUE #100: Would you consider implementing FromStr or serde for `Format`

ModProg:
closes #100


diwic:
Thanks! One could argue unknown strings should return an error instead but I don't know, maybe this is better.

Issue #103, opened 2023-05-26
Update to bitflags 2.0


JMS55:

diwic:
What's the motivation for this change?

JMS55:
Other crates Bevy uses like wgpu have already updated, so right now we have duplicate versions of bitflags in our dependency tree.

PR #107, opened 2023-07-25
Allow to build against bitflags 2.0

glandium:
Using bitflags 2.0 requires no changes, so it's fine to use 1.3.2 or newer 2.x versions.

Fixes: #103


glandium:
Mmm as this changes the traits derived for the public types, this doesn't quite work.

PR #113, opened 2023-10-13
add element lock/unlock bindings

chadmed:
allows an application to lock an element for exclusive control and unlock it

if we could have a minor release too that would be nice :)


diwic:
0.8.1 released now, enjoy!

PR #114, opened 2023-10-16
Update nix to 0.27

heiher:
Bump up nix to 0.27 to resolve compile issues for new targets (LoongArch/RISC-V).


diwic:
Maybe I should get rid of nix altogether. Seems to cause more churn than help these days :-/

PR #115, opened 2024-01-31
add unsafe_clone for PCM

bbarker:


diwic:
Hi and thanks for the PR,

It does not look right though. What problems would this solve that you can't solve by putting PCM into an `Arc<PCM>` or `Arc<Mutex<PCM>>`?

bbarker:
Yes, you are right - I actually had Arc<Mutex<...>> at a higher level, but forgot that I'd added a `[#derive(Clone)]` in between there and `PCM`. Thanks!

PR #116, opened 2024-02-13
Nuke nix

Jake-Shadle:
I noticed https://github.com/diwic/alsa-rs/pull/114#issuecomment-1769809723 when I was looking to see why the 0.8.1 version of this crate is using an older version of nix so figured removing nix would be fine. This crate has constantly been using slightly older versions of nix from other dependencies we use, and when I went to replace the use of nix found that the usage was incredibly minimal.

I only tested on linux and had the same test failures as before this change so don't think I broke anything, but I wasn't sure what the intended support for BSD-like targets is since this crate doesn't seem to state anything about them or have CI for them, but I did add support for them, though I failed to figure out to actually get eg. freebsd to build via cross.

### DIFF
--- a/src/direct/ffi.rs
+++ b/src/direct/ffi.rs
@@ -1,10 +1,10 @@
+//! Some definitions from the kernel headers
 
-// Some definitions from the kernel headers
+#![allow(non_camel_case_types)]
 
 // const SNDRV_PCM_MMAP_OFFSET_DATA: c_uint = 0x00000000;
 pub const SNDRV_PCM_MMAP_OFFSET_STATUS: libc::c_uint = 0x80000000;
 pub const SNDRV_PCM_MMAP_OFFSET_CONTROL: libc::c_uint = 0x81000000;
-
 
 pub const SNDRV_PCM_SYNC_PTR_HWSYNC: libc::c_uint = 1;
 pub const SNDRV_PCM_SYNC_PTR_APPL: libc::c_uint = 2;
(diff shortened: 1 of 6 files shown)


diwic:
Hi and thanks a lot for the PR!

I think the whole picture makes sense. A few questions:

1) Are you the owner of this code or are you copy/pasting it from some other project? Just to make sure there are no accidental license violations here.

2) I understand this will be a backwards incompatible change, is there a short summary of what a user is supposed to change during the upgrade?

src/error.rs
```diff
@@ -25,76 +25,191 @@ macro_rules! acheck {
...
 }
 
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "ALSA function '{}' failed with error '{}'", self.0, self.1)
+        write!(
+            f,
+            "ALSA function '{}' failed with error '{}'",
+            self.0,
+            desc(self.1)
```
diwic, line 116: Would probably make sense to write the number as well as the name here, so you don't end up with 'unknown errno' and no way to troubleshoot further.

src/error.rs
```diff
@@ -25,76 +25,191 @@ macro_rules! acheck {
...
+        libc::EHOSTDOWN => "Host is down",
+        libc::EHOSTUNREACH => "No route to host",
+        libc::ENOTSUP => "Operation not supported",
+        _ => "Unknown errno",
+    }
 }
 
+impl From<Error> for fmt::Error {
+    fn from(_: Error) -> fmt::Error {
+        fmt::Error
```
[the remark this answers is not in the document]
Jake-Shadle: https://github.com/diwic/alsa-rs/blob/db4bdd7f4e467148c091720589d217fdafb5d168/src/chmap.rs#L57 and https://github.com/diwic/alsa-rs/blob/db4bdd7f4e467148c091720589d217fdafb5d168/src/chmap.rs#L84 convert to a `fmt::Error`, normally I would have refactored this to avoid losing the error information but I wanted to keep the changes minimal.

Jake-Shadle:
1. No, the code was mostly copied piecemeal from nix, which is MIT licensed so I didn't think it would be problematic.
2. The only public change is that nix is no longer re-exported, so crates that depend on alsa that just use the re-export instead of their own dependency on nix would need to change, and the methods to retrieve the nix error directly are gone so if a crate was doing a direct conversion from the nix error to their own error type that would need to change.

src/direct/ffi.rs
```diff
@@ -25,54 +25,172 @@ pub type __kernel_off_t = libc::c_long;
...
 pub struct snd_pcm_sync_ptr {
-	pub flags: libc::c_uint,
-	pub s: snd_pcm_mmap_status_r,
-	pub c: snd_pcm_mmap_control_r,
+    pub flags: libc::c_uint,
+    pub s: snd_pcm_mmap_status_r,
+    pub c: snd_pcm_mmap_control_r,
+}
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
```
diwic, line 74: Add comment about mod ioctl_helpers copy-pasted from nix here

src/error.rs
```diff
@@ -25,76 +25,192 @@ macro_rules! acheck {
...
+            desc(self.1),
+            self.1,
+        )
     }
 }
 
-impl From<Error> for fmt::Error {
-    fn from(_: Error) -> fmt::Error { fmt::Error }
+fn desc(errno: i32) -> &'static str {
+    match errno {
```
diwic, line 123: Add comment about function copy-pasted from nix here

diwic (commented):
Okay, cool. Let's add comments on the bigger things you copy-pasted from nix so we leave a little credit. Then squash all commits into one and I think we're ready for merge.

diwic:
...and it's in!

Assuming you want an alsa 0.9 release with this code in it, is there any other dependencies I should update before I go ahead and release a 0.9?

Jake-Shadle:
Thanks! Nope, I don't believe so, libc has been 0.2 for forever and this crate already uses the latest bitflags, it was only nix that was the issue.

diwic:
Released now, enjoy!

Jake-Shadle:
Thanks for the quick turnaround!

Issue #118, opened 2024-03-02
PCM::direct_mmap_capture errors with ENOTTY on ioctl


tranzystorekk:
After updating to version 0.9.0 I get an ENOTTY error when attempting to call `PCM::direct_mmap_capture()`:

```
Error: ALSA function 'SNDRV_PCM_IOCTL_CHANNEL_INFO' failed with error 'Not a typewriter (25)'
```

The environment is:

Void Linux (vmware guest)
Linux 6.6.18
Vmware Workstation Player 17.5.0

diwic:
Interesting. Are you sure the difference is between two different alsa-rs versions and not something else in the environment (e g, not all drivers/devices support mmap)?

tranzystorekk:
I'm pretty sure, this reproduces when I update this project from 0.8.1 to 0.9.0: https://codeberg.org/tranzystorekk/spectroscope, here's where I call the API: https://codeberg.org/tranzystorekk/spectroscope/src/commit/376cfbf1151a3d77615b20822965fb728ea5ffa9/src/capture.rs#L32

For an overview, the program attempts to open an alsa loopback device to read audio for visualizing.

diwic:
@Jake-Shadle Hi, it's probably a regression from your "nuking nix" PR, would you mind having a look at this bug? Thanks!

Jake-Shadle:
Sorry about that, fixed in #119

tranzystorekk:
Just tested with the PR commit and the issue is fixed, thank you @Jake-Shadle ❤️

PR #119, opened 2024-03-05
Fix ioctl bug on linux

Jake-Shadle:
There was an off-by-one error (13 instead of 14) for the SIZEBITS value that caused the ioctl code to be screwed up. 

Resolves: #118


flosse:
@diwic would you mind to publish a new version with this fix?

diwic:
@flosse 0.9.1 has been released now, enjoy!

flosse:
Thank you!

PR #127, opened 2024-11-09
Add missing chmap positions

horazont:
These four positions were missing in the rust definition of the enum. No idea why. However, that made it impossible to construct any standard surround channel map (which need rear-right (4.0 onward), LFE (2.1 onward), front-center (5.0 onward), or side-left (7.1 onward)).


diwic:
Thanks! I have no idea why they were missing either.

diwic:
Let me know if you need a release.

horazont:
@diwic Thanks! I don't need a release for now, I'm just toying around so I can depend on master.

Issue #128, opened 2024-11-22
Memory leak alsa::pcm::PCM::open


druskus20:
I first found out this issue working with `cpal`. (https://github.com/RustAudio/cpal/issues/928)

I've reproduced this issue with this repo's `pcm_record` example. (I slightly modified it to make it not infinitely loop).

```rs
fn read_loop(pcm: &PCM) -> Result<(), Error> {
    let io = pcm.io_i16()?;
    let mut buf = [0i16; 8192];
    for i in 0..10 {
        // Block while waiting for 8192 samples to be read from the device.
        assert_eq!(io.readi(&mut buf)?, buf.len());
        let r = rms(&buf);
        println!("RMS: {:.1} dB", r);
    }
    Ok(())
}
```

Running: 
```
 valgrind -s --leak-check=full --show-leak-kinds=all ./target/debug/examples/pcm_record
```

Shows: 

```
==1331522== 16,616 bytes in 1 blocks are still reachable in loss record 149 of 149
==1331522==    at 0x484D953: calloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==1331522==    by 0x5159CE8: ??? (in /usr/lib/x86_64-linux-gnu/libpipewire-0.3.so.0.1005.0)
==1331522==    by 0x5174096: ??? (in /usr/lib/x86_64-linux-gnu/libpipewire-0.3.so.0.1005.0)
==1331522==    by 0x515A818: pw_init (in /usr/lib/x86_64-linux-gnu/libpipewire-0.3.so.0.1005.0)
==1331522==    by 0x50BA67B: _snd_pcm_pipewire_open (in /usr/lib/x86_64-linux-gnu/alsa-lib/libasound_module_pcm_pipewire.so)
==1331522==    by 0x48CB6CD: ??? (in /usr/lib/x86_64-linux-gnu/libasound.so.2.0.0)
==1331522==    by 0x48CBD6D: ??? (in /usr/lib/x86_64-linux-gnu/libasound.so.2.0.0)
==1331522==    by 0x48FB37E: snd_pcm_open (in /usr/lib/x86_64-linux-gnu/libasound.so.2.0.0)
==1331522==    by 0x122614: alsa::pcm::PCM::open (error.rs:21)
==1331522==    by 0x122510: alsa::pcm::PCM::new (pcm.rs:147)
==1331522==    by 0x120AAC: pcm_record::start_capture (pcm_record.rs:7)
[...]

PR #129, opened 2024-11-22
Fix Memory leak in alsa::pcm::PCM::open (?)

druskus20:
This PR attempts to fix https://github.com/diwic/alsa-rs/issues/128.

It does so by calling `alsa::snd_config_update_free_global();`. 

I am no alsa expert, so here is some more info on the topic:

* https://alsa-project.org/alsa-doc/alsa-lib/group___config.html#ga51dbc0ff1d9b34d753706d09d768116b
*  https://github.com/alsa-project/alsa-lib/blob/master/MEMORY-LEAK: `"The biggest reported leak is that the global configuration is cached for next usage."` 

Maybe there is a better place to free the global configuration, if caching it for successive usage is of interest?

## System 

I am using alsa `1.2.11`, on Ubuntu 24.04  6.8.0-48-generic.

```
dpkg -l | grep libasound2
ii  libasound2-data                                1.2.11-1build2                           all          Configuration files and profiles for ALSA drivers
ii  libasound2-dev:amd64                           1.2.11-1build2                           amd64        shared library for ALSA applications -- development files
ii  libasound2-plugins:amd64                       1.2.7.1-1ubuntu5                         amd64        ALSA library additional plugins
ii  libasound2t64:amd64                            1.2.11-1build2                           amd64        shared library for ALSA applications
```

## Test

To test this PR I have created a minimal example:

```rs
use alsa::pcm::*;
use alsa::Direction;

fn main() {
    let _ = PCM::new("default", Direction::Capture, false).unwrap();
    // PCM::drop is called
}
```
And ran it as follows:

```bash
cargo build --release --example minimal && valgrind -s --leak-check=full  ./target/release/examples/minimal
```

### Before applying this PR: 

```
==1453338==
==1453338== LEAK SUMMARY:
==1453338==    definitely lost: 1,560 bytes in 22 blocks
==1453338==    indirectly lost: 1,670 bytes in 12 blocks
==1453338==      possibly lost: 76,900 bytes in 2,360 blocks
==1453338==    still reachable: 37,113 bytes in 68 blocks
==1453338==         suppressed: 0 bytes in 0 blocks
==1453338== Reachable blocks (those to which a pointer was found) are not shown.
==1453338== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==1453338==
==1453338== ERROR SUMMARY: 114 errors from 114 contexts (suppressed: 0 from 0)
```

### After applying this PR

```
==1454406==
==1454406== LEAK SUMMARY:
==1454406==    definitely lost: 18,185 bytes in 24 blocks
==1454406==    indirectly lost: 2,226 bytes in 21 blocks
==1454406==      possibly lost: 0 bytes in 0 blocks
==1454406==    still reachable: 16,224 bytes in 43 blocks
==1454406==         suppressed: 0 bytes in 0 blocks
==1454406== Reachable blocks (those to which a pointer was found) are not shown.
==1454406== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==1454406==
==1454406== ERROR SUMMARY: 11 errors from 11 contexts (suppressed: 0 from 0)
```

## Leftover leaks

As you can see, some leaks still remain. And in fact, **definitely lost** has increased.

This is what valgrind has to say: 

```
==1454406== 17,172 (16,616 direct, 556 indirect) bytes in 1 blocks are definitely lost in loss record 38 of 38
==1454406==    at 0x484D953: calloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==1454406==    by 0x5159CE8: ???
==1454406==    by 0x5174096: ???
==1454406==    by 0x515A818: ???
==1454406==    by 0x50BA67B: ???
==1454406==    by 0x48CB6CD: ??? (in /usr/lib/x86_64-linux-gnu/libasound.so.2.0.0)
==1454406==    by 0x48CBD6D: ??? (in /usr/lib/x86_64-linux-gnu/libasound.so.2.0.0)
==1454406==    by 0x48FB37E: snd_pcm_open (in /usr/lib/x86_64-linux-gnu/libasound.so.2.0.0)
==1454406==    by 0x11C796: alsa::pcm::PCM::new (in /home/drusk/alsa-rs-pr/target/release/examples/minimal)
==1454406==    by 0x11C5C2: minimal::main (in /home/drusk/alsa-rs-pr/target/release/examples/minimal)
==1454406==    by 0x11C582: std::sys::backtrace::__rust_begin_short_backtrace (in /home/drusk/alsa-rs-pr/target/release/examples/minimal)
==1454406==    by 0x11C578: std::rt::lang_start::{{closure}} (in /home/drusk/alsa-rs-pr/target/release/examples/minimal)
```

I found this issue, referencing something similar, pointing to a possible leak in libpulse.so.0.21.1.
 
https://github.com/alsa-project/alsa-lib/issues/93

It seems like it does not happen on every system though, I haven't been able to test different versions of alsa, might get around to it at some point.

### DIFF
--- a/src/pcm.rs
+++ b/src/pcm.rs
@@ -340,7 +340,10 @@ impl PCM {
 }
 
 impl Drop for PCM {
-    fn drop(&mut self) { unsafe { alsa::snd_pcm_close(self.0) }; }
+    fn drop(&mut self) { unsafe {
+        alsa::snd_pcm_close(self.0);
+        alsa::snd_config_update_free_global();
+    }; }
 }
 
 


druskus20:
I actually got a similar example in C, to verify that the **leftover leaks** are still there in my system. 
I compiled it with both my libasound system library (1.2.11) and the leaks seem to still be there. 


```c
#include <alsa/asoundlib.h>
#include <stdio.h>

int main() {
  snd_pcm_t *handle;
  int err;
  err = snd_pcm_open(&handle, "default", SND_PCM_STREAM_PLAYBACK,
                     SND_PCM_NONBLOCK);
  printf("snd_pcm_open: %d\n", err);
  err = snd_pcm_close(handle);
  printf("snd_pcm_close: %d\n", err);
  snd_config_update_free_global();
  return 0;
}
```


TODO: 

Try different versions of pipewire / pulse stuff.

diwic:
Hi and thanks for the PR!

I think the first step would be to have a new `config` module with a `update_free_global` function. That way people can choose for themselves whether they care about this leak or not. I guess most would probably prefer the slight speed increase of keeping the cache between calls to "open", given that you're the first to care about this. 

Would you like to make a PR with such a module?

diwic:
I'm asking for a new module because this is not specific to PCM. Opening ctls or cards would likely result in the same leak.

druskus20:
Yes, making a module sounds reasonable. 
I might give it a crack but at the moment I don't have a lot of spare time - and I'm not familiar with Alsa. Though it seems easy enough. 

Update, opening a device such as `hw:CARD=Generic,DEV=7` (non pipewire) with my C example results in now leaks, but the PCM minimal example still produces other kinds of leaks such as:

```
==18547== 6,624 bytes in 92 blocks are possibly lost in loss record 108 of 108
==18547==    at 0x484D953: calloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==18547==    by 0x48B3D92: ??? (in /usr/lib/x86_64-linux-gnu/libasound.so.2.0.0)
==18547==    by 0x48C45B4: ??? (in /usr/lib/x86_64-linux-gnu/libasound.so.2.0.0)
==18547==    by 0x48C48B9: ??? (in /usr/lib/x86_64-linux-gnu/libasound.so.2.0.0)
==18547==    by 0x48C4D24: ??? (in /usr/lib/x86_64-linux-gnu/libasound.so.2.0.0)
==18547==    by 0x48C4D24: ??? (in /usr/lib/x86_64-linux-gnu/libasound.so.2.0.0)
==18547==    by 0x48C5289: ??? (in /usr/lib/x86_64-linux-gnu/libasound.so.2.0.0)
==18547==    by 0x48C558B: ??? (in /usr/lib/x86_64-linux-gnu/libasound.so.2.0.0)
==18547==    by 0x48C572E: ??? (in /usr/lib/x86_64-linux-gnu/libasound.so.2.0.0)
==18547==    by 0x48D06EB: snd_config_hook_load (in /usr/lib/x86_64-linux-gnu/libasound.so.2.0.0)
==18547==    by 0x48C9788: ??? (in /usr/lib/x86_64-linux-gnu/libasound.so.2.0.0)
==18547==    by 0x48C9CB4: ??? (in /usr/lib/x86_64-linux-gnu/libasound.so.2.0.0)
```

I haven't looked into why yet. I thought they would be identical but seems like alsa_rs is doing something more

druskus20:
Thanks for the fix! - Closing this since it's already fixed. 
The pipewire leaks likely do not belong here.

PR #130, opened 2024-12-13
Add ElemList

RavuAlHemio:
This allows enumerating control values on a device.

### DIFF
--- a/examples/ctl_list.rs
+++ b/examples/ctl_list.rs
@@ -1,0 +1,29 @@
+//! Example that enumerates controls for a device.
+
+use alsa::Card;
+use alsa::card::Iter;
+use alsa::ctl::Ctl;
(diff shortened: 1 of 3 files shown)


diwic:
This looks like a well written patch, thanks! My only thought is that one could avoid that extra call to `snd_ctl_elem_list_malloc` during initialization. 

And maybe add some kind of iterator, maybe one that outputs `ElemId` but I'm not sure if that's a good idea or not. 

Anyhow this patch is good enough to be merged as-is and if you want to write follow-up PRs with my suggestions (or your own) you're welcome to do so :-)

Let me know if you need a release. Thanks!

RavuAlHemio:
Thanks for the review and merge; I'll contribute additional PRs with your suggestions.

PR #131, opened 2024-12-14
ElemList: avoid heap allocation for buffer used to query size

RavuAlHemio:
(I hope I'm doing this right... comments are welcome!)


diwic:
Hmm, this is not exactly what I meant. You could have elem_list_new take a `snd_ctl_t*`...either that, or an FnOnce closure that returns the count.

RavuAlHemio:
Uhh, I'm afraid I don't follow. Do you have a similar example in the existing code?

diwic:
Like this https://github.com/diwic/alsa-rs/commit/2219abbb364e8b063af70db469652476fac39400

PR #133, opened 2025-04-18
fix: Add missing {S20,U20}{LE,BE} (20 bits for 4 bytes align)

yhx-12243:


diwic:
Hi, I'm curious if this is a format that you've found in the wild and need to support or it's more about completeness?

Also it seems this causes the test to fail, can you correct it an resubmit?

yhx-12243:
Basically it's about completeness since https://github.com/diwic/alsa-sys/blob/2fa68345010523d76e66c5f2377324fbfd6d124d/src/generated.rs#L964-L971 are only four formats that this crate does not provide.

Also I'm writing an application to match the fields for completeness and found that there are no these four fields.

PR #134, opened 2025-07-11
Add unsafe `readn` and` writen`

anas2660:
While actual hardware mostly only supports interleaved buffers, it can make sense to have available in some cases.
For example, using pipewire with pipewire-alsa and capturing data from external jack clients using non-interleaved buffers.


diwic:
It does not make sense for writei to take `*const` and writen to take `*mut`. This is probably a mistake on the alsa-lib side? 

`writen(&self, bufs: &[*const S], frames: usize)` would make more sense, or am I missing something?

diwic:
Thanks!

Issue #138, opened 2025-09-17
Handle `stderr` when opening device from alsa


dsgallups:
Relevant issues: 
- https://github.com/RustAudio/cpal/issues/605
- https://github.com/CorvusPrudens/bevy_seedling/issues/51

Because `alsa::snd_pcm_open` prints to `stderr`, `acheck!` doesn't capture these values.

diwic:
Hi and thanks for the PR!

Alsa-lib has a way of redirecting output from stderr to a custom error handler, see `snd_lib_error_set_handler`. No one has asked for it before now, so I don't think I've implemented anything in that area. 

Maybe we can use that instead? That said, callbacks from c is no fun because of Rust's ability to panic...

diwic:
Okay, so I found it easier to implement myself than to tell someone else how to do it, so now you can call `Output::local_error_handler` to redirect alsa-lib's errors there.

If you still get errors from jack, that is outside alsa-lib's control unfortunately.

dsgallups:
Thank you!!

dsgallups:
@diwic, Thank you for cleaning up a lot of the errors appropriately! Do you have any guidance on how I would be able to capture errors from jack server? 

i.e.

```console
jack server is not running or cannot be started
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
```

It would be nice to match on these errors in a more rusty way. As you had mentioned, it's outside the scope of alsa-lib, so I was wondering if this is something I should look into for `alsa` itself.

diwic:
@dsgallups 
There are two references to stderr here: https://github.com/alsa-project/alsa-plugins/blob/master/jack/pcm_jack.c that should call some SND_ERROR macro instead.

Maybe vaccuum alsa-lib and alsa-plugins and jack2 repos for references to stderr and submit patches upstream?

PR #139, opened 2025-09-17
fix: catch `stderr` when opening device

dsgallups:
related issue: #138 

I have taken some of the solution code here 
https://github.com/RustAudio/cpal/issues/605#issuecomment-1465065005.
However, I have made adjustments to restore `stderr` after the device has been returned.

I have added comments for suggestions on how errors should be handled in this case, as I am a bit naive about the internals of alsa. Please let me know if this isn't suitable, as I would like to be able to handle these errors in a more rusty way. Happy to do whatever!

Thank you!

### DIFF
--- a/src/pcm.rs
+++ b/src/pcm.rs
@@ -361,18 +361,20 @@ impl poll::Descriptors for PCM {
     }
 }
 
+// Keeping imports/code around for suggestions on handling this case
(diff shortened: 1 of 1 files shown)


diwic:
Maybe it is possible to use `snd_output_t` for this somehow...

PR #141, opened 2025-09-23
`no-std` feature flag.

### LINKED ISSUE #140: no_std feature flag
Looking over the code, it seems to me we could have a `no-std` feature flag while retaining most (if not all) of the crate's functionality. We would have to use an `extern alloc` crate for the `Vec` and `String` implementations, but otherwise it appears to me everything would be fine.

We can also create another `alloc` feature, gating just the functions that use `Vec` and `String`s behind it.

I would be willing to implement it myself if the `alsa-rs` maintainers aren't against it. A `no-std` flag would primarily make this crate more amenable for use in embedded devices, or for people who want to use it in otherwise memory constrained environments. There may be more use-cases I am unaware of.

----
EDIT: I've actually already finished a first implementation, and can open a PR if this is accepted. There are a lot of changes:

```
➜ g d origin/master --stat
  .github/workflows/rust.yml |    2 +
  Cargo.toml                 |    4 +
  src/card.rs                |   10 +-
  src/chmap.rs               |   15 +-
  src/config.rs              |    7 +-
  src/ctl_int.rs             |   14 +-
  src/device_name.rs         |   13 +-
  src/direct/asound_ioctl.rs | 2234 +++++++++++++++++++++++++++++++++++++++++++++++++++---------------------------------------------------
  src/direct/ffi.rs          |    4 +-
  src/direct/pcm.rs          |   34 +-
  src/error.rs               |   25 +-
  src/hctl.rs                |   27 +-
  src/io.rs                  |   13 +-
  src/lib.rs                 |    7 +
  src/mixer.rs               |   25 +-
  src/pcm.rs                 |   79 ++--
  src/poll.rs                |    6 +-
  src/rawmidi.rs             |   26 +-
  src/seq.rs                 |   38 +-
  19 files changed, 1327 insertions(+), 1256 deletions(-)
```

However, many of them are trivial (changing `std::` to `core::`). And I've separated them in multiple commits so that they are easier to review.

If this is not accepted, I can still use it for my personal projects, so it's not that big of a deal.

LGFae:
Closes #140.

I tried dividing it in multiple commits to make it easier to review.

First two commits should be very straightforward. The first was literally done with with a few  `find -exec sed` commands.

The third is where we had to change a few things to make them work in both `std` and `no-std` environments. In particular, we changed `last_os_error` to reading `errno` directly. My understanding is that those should be equivalent in unix systems, which is all ALSA cares about.

The fourth commit adds a check for the `no-std` flag to CI.

Going forward, supporting this flag should be relatively easy: use `core::` and `alloc::` instead of `std` in as many places as possible. In particular, this includes using `alloc::vec::Vec` and `alloc::string::String` instead of the containers from `std` (the implementation is actually the same, `std` just re-exports them). If there is ever a place where that cannot be done, if should be notated with `#[cfg(not(feature = "no-std"))]`. We have to do this in only 3 locations currently: two `std::io::{Read, Write}` implementations and one use of a `thread_local!` global.

---
EDIT: one more thing is that you will have to manually declare `extern crate std` in tests functions to use the `std` crate.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -17,6 +17,7 @@ pub struct Error(&'static str, c_int);
 
 pub type Result<T> = ::core::result::Result<T, Error>;
 
+#[cfg(not(feature = "std"))]
 extern "C" {
     pub(crate) static errno: c_int;
 }
@@ -70,10 +71,18 @@ impl Error {
     }
 
     pub fn last(func: &'static str) -> Error {
-        Self(
-            func,
-            unsafe {errno},
-        )
+        #[cfg(feature = "std")] {
+            Self(
+                func,
+                std::io::Error::last_os_error().raw_os_error().unwrap_or_default(),
+            )
+        }
+        #[cfg(not(feature = "std"))] {
+            Self(
+                func,
+                unsafe {errno},
+            )
+        }
     }
 
     pub fn unsupported(func: &'static str) -> Error {
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -44,7 +44,14 @@ impl Descriptors for pollfd {
 pub fn poll(fds: &mut[pollfd], timeout: i32) -> Result<usize> {
     let r = unsafe { libc::poll(fds.as_mut_ptr(), fds.len() as libc::nfds_t, timeout as libc::c_int) };
     if r >= 0 { Ok(r as usize) } else {
-        from_code("poll", -unsafe {super::error::errno}).map(|_| unreachable!())
+        #[cfg(feature = "std")]
+        { from_code(
+            "poll",
+            -std::io::Error::last_os_error().raw_os_error().unwrap_or_default()).map(|_| unreachable!())
+        }
+        #[cfg(not(feature = "std"))]
+        { from_code("poll", -unsafe {super::error::errno}).map(|_| unreachable!()) }
+        
     }
 }
 


src/direct/pcm.rs
```diff
@@ -20,12 +20,12 @@ For an example of how to use this mode, look in the "synth-example" directory.
 */
 
 use libc;
-use std::{mem, ptr, fmt, cmp};
+use core::{mem, ptr, fmt, cmp};
 use crate::error::{Error, Result};
-use std::os::unix::io::RawFd;
+use core::ffi::c_int;
```
diwic, line 25: As for RawFd. Would it be possible to declare it somewhere like

```
#feature(std)
type RawFd = std::os::unix::io::RawFd;
#feature(no-std)
type RawFd = core::ffi::c_int;
```

...and then use that definition of `RawFd` everywhere else?
Makes the code more readable I believe.

src/lib.rs
```diff
@@ -126,6 +131,8 @@ pub use crate::mixer::Mixer as Mixer;
 pub mod seq;
 pub use crate::seq::Seq as Seq;
 
+/// The io module uses the thread_local macro. There is no equivalent of it in `core` or `alloc`,
```
diwic, line 134: Was this comment left behind? You do compile the io module either way.

src/io.rs
```diff
@@ -39,6 +40,7 @@ impl Output {
     ///
     /// Sometimes alsa-lib writes to stderr, but if you prefer, you can write it here instead.
     /// Should you wish to empty the buffer; just call local_error_handler again and drop the old instance.
+    #[cfg(feature = "std")]
```
diwic, line 43: Do you think we need to write in the comment to this function that it's unavailable in no-std mode or is this obvious when you see it in docs.rs later?

src/error.rs
```diff
@@ -12,9 +13,13 @@ use std::{fmt, str};
 /// An Error is also returned in case ALSA returns a string that
 /// cannot be translated into Rust's UTF-8 strings.
 #[derive(Clone, PartialEq, Copy)]
-pub struct Error(&'static str, i32);
+pub struct Error(&'static str, c_int);
 
-pub type Result<T> = ::std::result::Result<T, Error>;
+pub type Result<T> = ::core::result::Result<T, Error>;
+
+extern "C" {
```
diwic, line 20: I think we should have different implementations for std and no-std mode. So that the errno import is only done in no-std mode, keeps the std mode cleaner. Like

```
#feature(std)
std::io::Error::last_os_error().raw_os_error().unwrap_or_default()
#feature(no-std)
unsafe { errno }
```

src/direct/pcm.rs
```diff
@@ -615,9 +619,9 @@ fn playback_to_plughw_mmap() {
     assert_eq!(m.appl_ptr(), 0);
     assert_eq!(m.hw_ptr(), 0);
 
-    println!("{:?}", m);
+    std::println!("{:?}", m);
     let mut i = (0..(m.buffer_size() * 2)).map(|i|
-        (((i / 2) as f32 * 2.0 * ::std::f32::consts::PI / 128.0).sin() * 8192.0) as i16);
+        (((i / 2) as f32 * 2.0 * ::core::f32::consts::PI / 128.0).sin() * 8192.0) as i16);
```
diwic, line 624: Is this change needed as the std crate is imported?

diwic (commented):
Hi and thanks for your work!

I think it looks good for the most part, I have a few questions/thoughts, see review comments. Thanks!

LGFae:
```
---- rawmidi::print_rawmidis stdout ----

thread 'rawmidi::print_rawmidis' (9961) panicked at src/rawmidi.rs:212:75:
called `Result::unwrap()` on an `Err` value: ALSA function 'snd_ctl_rawmidi_next_device' failed with error 'Not a typewriter (25)'
```

PR #152, opened 2026-06-11
Flesh out initial UMP support (MIDI 2.0)

abique:


src/ump.rs
```diff
@@ -0,0 +1,147 @@
...
+                ptr::null_mut()
+            },
+            name.as_ptr(),
+            flags
+        ))
+        .map(|_| Ump(h))
+    }
+
+    pub fn info(&self) -> Result<Info> {
+        Info::new().and_then(|i| acheck!(snd_ump_rawmidi_info(self.0, i.0)).map(|_| i))
```
diwic, line 55: Per convention, function should be named `rawmidi_info` instead of `info`

src/ump.rs
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    pub fn status(&self) -> Result<Status> {
+        Status::new().and_then(|i| acheck!(snd_ump_rawmidi_status(self.0, i.0)).map(|_| i))
```
diwic, line 59: Per convention, function should be named `rawmidi_status` instead of `status`

diwic (commented):
Hi and thanks for looking into this! Fun to see new stuff appearing in alsa and these bindings.

src/ump.rs
```diff
@@ -0,0 +1,147 @@
...
+impl Ump {
+    /// Wrapper around open that takes a &str instead of a &CStr
+    pub fn new(name: &str, dir: Direction, nonblock: bool) -> Result<Self> {
+        Self::open(&CString::new(name).unwrap(), dir, nonblock)
+    }
+
+    pub fn open(name: &CStr, dir: Direction, nonblock: bool) -> Result<Ump> {
+        let mut h = ptr::null_mut();
+        let flags = if nonblock { 2 } else { 0 }; // FIXME: alsa::SND_RAWMIDI_NONBLOCK does not exist in alsa-sys
+        acheck!(snd_ump_open(
```
diwic, line 37: Maybe we also want an `open` variant that returns both a playback and record device?

abique: I think it is better to call open twice.

src/ump.rs
abique: Yes, though I'm not sure if there's a benefit in specifying `rawmidi` here?

abique: I'll add rawmidi here as requested.

src/ump.rs
diwic: It seems like we do the same for rawmidi, so I'll let it pass as is.

abique:
@diwic thanks you for your fast reaction to the PR :+1:

PR #155, opened 2026-06-19
Initial support for snd_rawmidi_params_t

abique:
It is unclear to me how to set the clock type when using ump.


abique:
Ah, I've got it.

src/rawmidi.rs
```diff
@@ -70,6 +70,48 @@ impl Info {
...
+    pub fn get_avail_min(&self) -> usize {
+        unsafe { alsa::snd_rawmidi_params_get_avail_min(self.0) as usize }
+    }
+
+    pub fn get_no_active_sensing(&self) -> bool {
+        let v = unsafe { alsa::snd_rawmidi_params_get_no_active_sensing(self.0) };
+        v != 0
+    }
+
+    pub fn get_read_mode(&self) -> alsa::snd_rawmidi_read_mode_t {
```
diwic: Enums are not directly exposed; we create alsa_enums around them instead

src/rawmidi.rs
```diff
@@ -189,6 +231,31 @@ impl Rawmidi {
 
     #[cfg(feature = "std")]
     pub fn io(&self) -> IO<'_> { IO(self) }
+
+    pub fn params_set_avail_min(&self, params: &mut Params, val: usize)-> Result<()> {
```
diwic: It would make more sense to have params_set_avail_min in the params struct, not in this one.

src/ump.rs
```diff
@@ -108,6 +110,35 @@ impl Ump {
     pub fn nonblock(&mut self, nonblock: i32) -> Result<()> {
         acheck!(snd_ump_nonblock(self.0, nonblock)).map(|_| ())
     }
+
+    fn rawmidi(&self) -> *mut alsa::snd_rawmidi_t {
+        unsafe { snd_ump_rawmidi(self.0) }
+    }
```
diwic: I think we should have something like `pub fn rawmidi(&self) -> &Rawmidi` here.

src/ump.rs
```diff
@@ -108,6 +110,35 @@ impl Ump {
...
+
+    pub fn params_set_avail_min(&self, params: &mut Params, val: usize)-> Result<()> {
```
diwic: Belongs to Params struct. So you'll use it like `params.set_avail_min(ump.rawmidi(), val)`

src/ump.rs
```diff
@@ -108,6 +110,35 @@ impl Ump {
...
+
+    pub fn params_set_read_mode(&self, params: &mut Params, val: alsa::snd_rawmidi_read_mode_t) -> Result<()> {
+        acheck!(snd_rawmidi_params_set_read_mode(self.rawmidi(), params.0, val)).map(|_| ())
+    }
+
+    pub fn params_set_clock_type(&self, params: &mut Params, val: alsa::snd_rawmidi_clock_t) -> Result<()> {
+        acheck!(snd_rawmidi_params_set_clock_type(self.rawmidi(), params.0, val)).map(|_| ())
+    }
+
+    pub fn params_current(&self) -> Result<Params> {
```
diwic: Rename to rawmidi_params_current. This way our users know that this corresponds 1:1 to snd_ump_rawmidi_params_current in the alsa-lib API.

src/ump.rs
```diff
@@ -108,6 +110,35 @@ impl Ump {
...
+        let params = Params::new()?;
+        acheck!(snd_ump_rawmidi_params_current(self.0, params.0)).map(|_| params)
+    }
+
+    pub fn params(&mut self, params: &Params) -> Result<()> {
```
diwic: Rename to rawmidi_params

src/rawmidi.rs
abique: Done.

src/ump.rs
abique: Done.

src/ump.rs
abique: Done.

src/ump.rs
abique: Done.

src/ump.rs
abique: Done

src/rawmidi.rs
abique: Done.

PR #156, opened 2026-07-07
PCM: add open_with_flags()

abique:
Hey,

Here's a PR to be able to pass undocumented flags to `snd_pcm_open()`.

A useful flag is 256 to open the PCM for scanning even if the PCM is already in use.

Cheers,
Alex

### DIFF
--- a/src/pcm.rs
+++ b/src/pcm.rs
@@ -153,11 +153,11 @@ impl PCM {
 
     // Does not offer async mode (it's not very Rustic anyway)
     pub fn open(name: &CStr, dir: Direction, nonblock: bool) -> Result<PCM> {
-        Self::open_with_flags(name, dir, nonblock, 0)
+        unsafe { Self::open_with_flags(name, dir, nonblock, 0) }
     }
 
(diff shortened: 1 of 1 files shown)


src/pcm.rs
```diff
@@ -153,13 +153,18 @@ impl PCM {
 
     // Does not offer async mode (it's not very Rustic anyway)
     pub fn open(name: &CStr, dir: Direction, nonblock: bool) -> Result<PCM> {
+        Self::open_with_flags(name, dir, nonblock, 0)
+    }
+
+    /// Same as open, except you can pass additional and undocumented flags at your own risk.
+    pub fn open_with_flags(name: &CStr, dir: Direction, nonblock: bool, user_flags: i32) -> Result<PCM> {
```
diwic: Should it be an `unsafe` function then, if it is "at your own risk"?

diwic: I want to know the risks :-)

abique: I can add that 🤝

PR #157, opened 2026-07-07
Fix compilation with time64-enabled libc

Gelbpunkt:
Similar to 62722db312fe078c4ab7a1f5c832e310af9d51eb, but for the rest of the codebase.


src/rawmidi.rs
```diff
@@ -256,10 +257,7 @@ impl Rawmidi {
     }
 
     pub fn tread(&self, buf: &mut [u8]) -> Result<(timespec, usize)> {
-        let mut timestamp: timespec = timespec {
-            tv_sec: 0,
-            tv_nsec: 0,
-        };
+        let mut timestamp: timespec = unsafe { zeroed() };
```
WhyNotHugo, line 260: You can use `timespec::default()` which is functionally equivalent without `unsafe`.

Gelbpunkt: I'm aware, but it is entirely safe to zero-initialize a timespec and this is already done elsewhere in the codebase

diwic: @WhyNotHugo timespec::default() is depending on the extra-traits feature of libc which might not be enabled.

